PR #9, opened 2021-11-29
Syntax highlight stub files

szepeviktor:
Like so https://github.com/szepeviktor/Valinor/blob/patch-1/stubs/Psr/SimpleCache/CacheInterface.php.stub

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,4 +11,4 @@
 /phpstan.neon.dist export-ignore
 /phpunit.xml.dist export-ignore
 
-*.php.stub linguist-language=PHP
+*.stub linguist-language=PHP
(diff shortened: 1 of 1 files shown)


romm:
Hi, thanks!

I didn't know about it, that's coming from https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes right?

Also, I see you changed from `*.php.stub` to `*.stub`, the first one didn't work?

szepeviktor:
> the first one didn't work?

Maybe I was lame. It worked with `*.stub` for a second try.
Double extensions look like a file from a phishing campaign! Please consider using `.stub`.

romm:
It's true that [PHPStan uses the `stub` extension as well](https://github.com/phpstan/phpstan-src/tree/447cd102d946a2d9883ac82793195275f54296a9/stubs), so let's be consistent indeed!

PR #10, opened 2021-11-29
Fix CI config

szepeviktor:
`8.1` is a floating point number in YAML

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -9,7 +9,7 @@ jobs:
     name: Mutation tests
     runs-on: ubuntu-latest
     env:
-      php-version: 8.1
+      php-version: '8.1'
       php-extensions: xdebug, yaml
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


szepeviktor:
[off] Die-hard YAML writers [use quotes](https://github.com/szepeviktor/debian-server-tools/blob/master/.travis.yml#L59) on every string.

szepeviktor:
@romm We don't need a trigger. Only that workflow is configured to run on PR-s.
https://github.com/CuyZ/Valinor/blob/396f64a5246ccfe3f6f6d3211bac7f542a9c7fc6/.github/workflows/mutation.yml#L2-L6
Would you like me to copy that to the other two?

romm:
I thought that the `push` option was sufficient (see below), because well… you obviously pushed your commits — maybe we need to add `pull_request` as well.

https://github.com/CuyZ/Valinor/blob/675e88cadfd5f4618feb7062fab6a0b4d0fd0801/.github/workflows/tests.yml#L1-L4

@Ocramius you are certainly more used to OSS contributions than I am, any input on this? 😛

szepeviktor:
> I thought that the `push` option was sufficient

This is a PR, I push in my fork.

Ocramius:
`push` is only for pushes on the source repo (or on the forks, where actions would need to be enabled too).

`pull_request` should be used for any CI actions that need to run in PRs: the `pull_request` trigger will not use any repository secrets, if you have configured any.

szepeviktor:
> any input on this? 😛

This is the mother of all GitHub Actions for PHP: https://github.com/ergebnis/php-package-template
(actually for a PHP package)
Please consider using/copying it.

romm:
Thank you both for the clarification, and for the link to the GitHub action repo, I'll probably take a look at some point.

In the meantime, would you mind adding the `pull_request` key yourself to see if workflow run correctly afterwards?

szepeviktor:
Done.

szepeviktor:
BTW These changes make test run twice if you open a PR from a branch of this repository.

romm:
> BTW These changes make tests run twice if you open a PR from a branch of this repository.

Yes that's what I thought, but I think that I'll have my own fork anyway so let's see how this goes. :)

Thanks!

PR #11, opened 2021-11-29
Add Node::hasChildren

szepeviktor:
When PHPStan says _Construct empty() is not allowed. Use more strict comparison._

### DIFF
--- a/src/Mapper/Printer/HtmlTreePrinter.php
+++ b/src/Mapper/Printer/HtmlTreePrinter.php
@@ -99,7 +99,7 @@ final class HtmlTreePrinter implements TreePrinter
     {
         $value = $node->isValid() ? $node->value() : null;
 
-        if (! empty($node->children())) {
+        if ($node->hasChildren()) {
             $details = 'details';
             $summary = 'summary';
         } else {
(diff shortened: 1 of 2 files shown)


src/Mapper/Printer/HtmlTreePrinter.php
```diff
@@ -99,7 +99,7 @@ public function print(Node $node): string
     {
         $value = $node->isValid() ? $node->value() : null;
 
-        if (! empty($node->children())) {
+        if ($node->hasChildren()) {
```
Ocramius, line 102: Tests are really really really needed for this sort of change :D

src/Mapper/Tree/Node.php
```diff
@@ -135,6 +135,11 @@ public function children(): array
         return $this->children;
     }
 
+    public function hasChildren(): bool
```
Ocramius, line 138: Tests are really really really needed for this sort of change :pray:

Ocramius (changes requested):

src/Mapper/Printer/HtmlTreePrinter.php
romm: Damn, this code should never have landed in the public repo :D

I'll probably delete this class anyway, and re-work on it later when other major issues are resolved.

szepeviktor: All right I close this repo.
Now I don't need to learn how to write tests 🙃

PR #32, opened 2021-12-06
feat: handle integer range type

romm:
Integer range can be used as follows:

```php
final class SomeClass
{
    /** @var int<42, 1337> */
    public int $intRange; // accepts any int between 42 and 1337

    /** @var int<-1337, 1337> */
    public int $negativeIntRange; // also works with negative values

    /** @var int<min, 1337> */
    public int $minIntRange; // `min` can be used…

    /** @var int<0, max> */
    public int $maxIntRange; // …as well as `max`
}
```

Note that `min` and `max` will check the range with PHP's internal
constants `PHP_INT_MIN` and `PHP_INT_MAX`.


src/Type/Parser/Exception/Scalar/ClassStringClosingBracketMissing.php
```diff
@@ -2,11 +2,13 @@
...
 declare(strict_types=1);
 
-namespace CuyZ\Valinor\Type\Parser\Exception;
+namespace CuyZ\Valinor\Type\Parser\Exception\Scalar;
 
+use CuyZ\Valinor\Type\Parser\Exception\InvalidType;
 use CuyZ\Valinor\Type\Type;
 use RuntimeException;
 
+// @todo move to \Scalar
```
stof, line 11: isn't this TODO done ?

romm: Indeed it is, thanks!

staabm:
positive-int and negative-int are internally also represented as IntegerRangeType, meaning I think this PR also adda support for those.
Didn‘t verify though

romm:
> positive-int and negative-int are internally also represented as IntegerRangeType, meaning I think this PR also adda support for those. Didn‘t verify though

Hi @staabm, actually `positive-int` and `negative-int` are already a feature, see:

https://github.com/CuyZ/Valinor/blob/4f561290b1ba71dc15f1b2806d2ea929429d5910/README.md?plain=1#L225-L229

PR #36, opened 2021-12-13
Failing test for array of union of objects

lookyman:
Hi. Readme says that the library can handle array of union of objects, but that doesn't seem to be true, as per this failing test. Please let me know if I'm doing anything wrong, and if not, do you need any help fixing?


romm:
Hi @lookyman, thank you for this very good contribution!

I have to give it some thoughts though; the documentation states that the mapper is able to infer an union of several interfaces, but this is not (yet?) the case for objects. There might be a manual solution, or maybe an automatic one. I'll try to take some time to work on it.

romm:
Hi @lookyman, with #56 your test is passing. Since I added a full testsuite for your case and other more complex ones, I will close this one.

Please give feedback if you encounter some issue with this new feature!

PR #40, opened 2021-12-17
Fixes #39 - Adds MySQL TIMESTAMP and PgSQL DATETIME formats.

### LINKED ISSUE #39: Example code does not work with library
I am trying to instantiate an object using this library, and using `DateTimeInterface` type hints as described in the example in the README. However, I get a CannotMapObject exception when I do. Does this library not yet support using a `DateTimeInterface` type hint for date times? Thanks.

brandonsavage:
I'm not sure I put these in the right place, but if there's a better place, let me know. 

I see this package being incredibly useful for automatically mapping data from databases into entities/value objects, so I hope that you'll consider this patch.

### DIFF
--- a/src/Mapper/Object/DateTimeObjectBuilder.php
+++ b/src/Mapper/Object/DateTimeObjectBuilder.php
@@ -21,6 +21,9 @@ use function is_string;
 
 final class DateTimeObjectBuilder implements ObjectBuilder
 {
+    public const DATE_MYSQL = 'Y-m-d H:i:s';
+    public const DATE_PGSQL = 'Y-m-d H:i:s.u';
+
     /** @var class-string<DateTime|DateTimeImmutable> */
     private string $className;
 
@@ -75,7 +78,7 @@ final class DateTimeObjectBuilder implements ObjectBuilder
         $formats = [
             DATE_ATOM, DATE_RFC850, DATE_COOKIE, DATE_RFC822, DATE_RFC1036,
             DATE_RFC1123, DATE_RFC2822, DATE_RFC3339, DATE_RFC3339_EXTENDED,
-            DATE_RFC7231, DATE_RSS, DATE_W3C,
+            DATE_RFC7231, DATE_RSS, DATE_W3C, self::DATE_MYSQL, self::DATE_PGSQL,
         ];
 
         foreach ($formats as $format) {
--- a/src/Mapper/Object/DateTimeObjectBuilder.php
+++ b/src/Mapper/Object/DateTimeObjectBuilder.php
@@ -76,9 +76,9 @@ final class DateTimeObjectBuilder implements ObjectBuilder
     private function tryAllFormats(string $value): ?DateTimeInterface
     {
         $formats = [
-            DATE_ATOM, DATE_RFC850, DATE_COOKIE, DATE_RFC822, DATE_RFC1036,
-            DATE_RFC1123, DATE_RFC2822, DATE_RFC3339, DATE_RFC3339_EXTENDED,
-            DATE_RFC7231, DATE_RSS, DATE_W3C, self::DATE_MYSQL, self::DATE_PGSQL,
+            self::DATE_MYSQL, self::DATE_PGSQL, DATE_ATOM, DATE_RFC850, DATE_COOKIE,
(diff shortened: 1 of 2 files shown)


src/Mapper/Object/DateTimeObjectBuilder.php
```diff
@@ -72,10 +75,10 @@
~
~    private function tryAllFormats(string $value): ?DateTimeInterface
~    {
         $formats = [
             DATE_ATOM, DATE_RFC850, DATE_COOKIE, DATE_RFC822, DATE_RFC1036,
```
romm, line 79: I think we could even add those on top of the list, there are probably more
chances these formats will be used more often compared to the others. WDYT?

brandonsavage: I think this is a greaet suggestion and I added commits to this goal.

romm:
Thank you very much for the contribution!

> I see this package being incredibly useful for automatically mapping data from databases into entities/value objects, so I hope that you'll consider this patch.

You don't rely on an ORM and do the hydration of your value objects yourself? I'm just a bit curious about how you use the library. Happy it helps though. 🙂

brandonsavage:
Typically I build my own domain model rather than using an ORM. The logic is pretty straightforward and very easy to test. In this case, I would be populating value and entity objects from raw data pulled from the database. 

Sample code:

```php
    public function findUserByEmail(string $email) : ?User
    {
        $select = $this->queryFactory->newSelect();
        $select->cols(['*'])->from('user_account')->where('email = :email');
        $result = $this->pdo->fetchOne($select, ['email' => $email]);
        if (!$result) {
            return null;
        }

        if ($result['last_login']) {
            $result['last_login'] = [
                'datetime' => $result['last_login'],
                'format' => 'Y-m-d H:i:s.u',
            ];
        }

        return (new MapperBuilder())->mapper()->map(
            User::class,
            $result
        );

    }
```

romm:
I see; I am also trying out [DDD](https://herbertograca.com/2019/06/05/reflecting-architecture-and-domain-in-code/) with a customer, but we still use Doctrine's ORM although we already thought about not using it at all (at least for read-models); we'll maybe reconsider this option.

Thanks again for the contribution. 🙂

Mediagone:
> Typically I build my own domain model rather than using an ORM. 
Domain model has nothing to do with the usage of an ORM or not.

The _Domain Model_ is the ensemble of all entities that fit your business requirements. Then, you have several ways to bind it with your persistence layer (database): you can use an ORM to do the work automatically or hydrate objects by yourself.

Notice that you still use a kind of ORM (in you example code), by requiring a "mapper" to build your object (an ORM being somehow a "mapper"). So what's the point to replace Doctrine ORM with this custom MapperBuilder? 
A simpler solution for manual hydration would be: `return User::fromArray($result);`

If the point to drop Doctrine ORM was performance, you should know that it's generally caused by a poor usage of the ORM rather than the ORM itself. You can, for example, use the `NEW` SQL operator to hydrate a DTO instead of entities.
( see: https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/dql-doctrine-query-language.html#new-operator-syntax )

Some useful links:
https://www.slideshare.net/guilhermeblanco/orm-dont-kill-your-db-developers-do
https://ocramius.github.io/doctrine-best-practices/#/

tests/Integration/Mapping/Type/DateTimeMappingTest.php
```diff
@@ -28,6 +29,8 @@ public function test_datetime_properties_are_converted_properly(): void
             'datetime' => '2012-12-21 13:37:42',
             'format' => 'Y-m-d H:i:s',
         ];
+        $mysqlDate = '2012-12-21 13:37:42';
+        $pgsqlDate = '2012-12-21 13:37:42.123456';
```
keichinger, line 33: I'm honestly wondering whether this test with these data is any useful. The current mapper's implementation is brute-forcing through various formats, using the first one that works.
The mechanism itself may be fine, though in this test each and every use-case is testing against the very same date time, potentially allowing/hiding false-positives where one format succeeds over the one that was supposed to be tested.

My proposal here would be to change each test-case to a different datetime to ensure that each test is actually doing what it's supposed to do.

romm: Very good advice, thank you for the review.

PR #50, opened 2021-12-28
feat!: allow mapping to any type 

romm:
Previously, the method `TreeMapper::map` would allow mapping only to an
object. It is now possible to map to any type handled by the library.

It is for instance possible to map to an array of objects:

```php
$objects = (new MapperBuilder())->mapper()->map(
    'array<' . SomeClass::class . '>',
    [/* … */]
);
```

For simple use-cases, an array shape can be used:

```php
$array = (new MapperBuilder())->mapper()->map(
    'array{foo: string, bar: int}',
    [/* … */]
);

echo $array['foo'];
echo $array['bar'] * 2;
```

This new feature changes the possible behaviour of the mapper, meaning
static analysis tools need help to understand the types correctly. An
extension for PHPStan and a plugin for Psalm are now provided and can be
included in a project to automatically increase the type coverage.


README.md
```diff
@@ -370,6 +408,91 @@ final class SomeClass
 }
 ```
 
+## Static analysis
+
+To help static analysis of a codebase using this library, an extension for
+[PHPStan] and a plugin for [Psalm] are provided. They enable these tools to
+understand better the behaviour of the mapper.
```
Mopolo, line 415: I think it should be written like this:

```suggestion
better understand the behaviour of the mapper.
```

PR #54, opened 2022-01-05
fix: README: Mapping error must be linked to a property

eigan:
'someValue' is not a key in `Node::children()` as used later in the example.

image

After change

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,10 +188,7 @@ the exception.
 ```php
 final class SomeClass
 {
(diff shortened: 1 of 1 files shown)


romm:
Hi @eigan, thank you for noticing this typo although the fix you're proposing is not really logical in my opinion.

With #52 merged, the example changed in the README and this typo is not relevant anymore, so I'm closing this PR.

Thanks anyway!

PR #58, opened 2022-01-07
misc!: mark classes with `@internal` or `@api` annotation

### LINKED ISSUE #4: Consider making most internal API `@internal` before moving to `1.0.0`
Currently, most of the classes in the package are public (no `@internal` declared), but there are loads of details inside the `Definition`, `Library`, `Mapper`, `Type` and `Utility` namespaces that are likely fit for being marked `@internal` from the start.

While that will indeed reduce the API that third-parties can rely upon, it will give you more runway to change the API if/when needed, and then to relax this boundary when you call a component stable enough for others to tinker with it.

romm:
This change filters the scope of the public API that is provided by this
library.

Any class or interface that is not explicitly marked with an `@api`
annotation should never be used outside this library — any change can
and will be made without taking breaking changes rules into account.

When a breaking change happens inside the public API scope, a major
version will be released; refer to https://semver.org for more
information.

A new PHPStan extension is now part of the quality assurance process,
ensuring that all classes/interfaces must provide either `@internal` or
`@api` annotation.

Fixes #4

### DIFF
--- a/qa/PHPStan/Extension/ApiAndInternalAnnotationCheck.php
+++ b/qa/PHPStan/Extension/ApiAndInternalAnnotationCheck.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CuyZ\Valinor\QA\PHPStan\Extension;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\InClassNode;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+use function str_starts_with;
+
+/**
+ * @implements Rule<InClassNode>
+ */
+final class ApiAndInternalAnnotationCheck implements Rule
+{
+    public function getNodeType(): string
+    {
+        return InClassNode::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        $reflection = $scope->getClassReflection();
+
+        if (! $reflection) {
+            return [];
+        }
+
+        if ($reflection->isAnonymous()) {
+            return [];
+        }
+
+        if (str_starts_with($reflection->getName(), 'CuyZ\Valinor\Tests')) {
+            return [];
+        }
+
+        if (str_starts_with($reflection->getName(), 'CuyZ\Valinor\QA')) {
+            return [];
+        }
+
+        if (! preg_match('/@api|internal\s+/', $reflection->getResolvedPhpDoc()?->getPhpDocString() ?? '')) {
+            return [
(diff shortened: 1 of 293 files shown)


qa/PHPStan/Extension/ApiAndInternalAnnotationCheck.php
```diff
@@ -0,0 +1,55 @@
...
+
+        if (str_starts_with($reflection->getName(), 'CuyZ\Valinor\Tests')) {
+            return [];
+        }
+
+        if (str_starts_with($reflection->getName(), 'CuyZ\Valinor\QA')) {
+            return [];
+        }
+
+        if (! preg_match('/@api|internal\s+/', $reflection->getResolvedPhpDoc()?->getPhpDocString() ?? '')) {
```
jdreesen, line 45: Shouldn't the regex be `/@(api|internal)\s+/` instead?

romm: You're right, thanks! Too bad I just merged it — wanna do a quick contribution to the repo? 😁

jdreesen: Sure: #61

kelunik:
@romm [`Reflection`](https://github.com/CuyZ/Valinor/blob/810c1abc6f3fe3cef17b6c68e38030b15c9e51f7/src/Utility/Reflection/Reflection.php#L111) is now internal and aliases still needed to be resolved when using it anyway. Is there anything in the public API now to get the return type or docblock return type from reflection?

romm:
Hi @kelunik, this is indeed a wanted change as this class is not intended to be part of the public API that needs to be maintained properly.

May I ask in what kind of scope you need this feature?

kelunik:
@romm Yeah, it's totally fine for this to be an internal API. I'd like to write an automatic mapper based on interfaces and [`ProxyManager`](https://github.com/Ocramius/ProxyManager), e.g. for database results or HTTP responses, which reads the expected type from reflection and / or the documented return type of the interface, and I thought I might be able to use some API of Valinor for that apart from using it for the mapping itself.

romm:
I see. Unfortunately I do not plan to make this API public, although it already crossed my mind to export all of this into another repository. Time will tell, but in the meantime I'm afraid you wont be able to use Valinor for this kind of things, sorry. 🙂

kelunik:
Don't worry, I'll just copy the relevant parts I need (respecting the license of course). Valinor will still be useful for doing all the heavy lifting under the hood. Thanks!

romm:
Glad it helps. :)

PR #59, opened 2022-01-11
Better type inference, improved test coverage throughout locked dependencies, automated dependency upgrades

Ocramius:
This patch is mostly raised by the upgrade of `0.3.0` to `0.4.0` breaking BC (expected, but preventable).

In practice, Valinor is probably going to be used because of `TreeMapper#map(class-string<T>, mixed): T` in the grand majority of cases.

To cover this scenario, and prevent regressions, I introduced:

 * testing against **locked** dependencies, so that we can pinpoint tiny upgrades that break `cuyz/valinor` as they come up
 * automated dependency **widening** (I would prefer bumping, but that's really your decision)
 * `vimeo/psalm` type inference testing
 * better types around `TreeMapper#map()`, so that it works without coupling `cuyz/valinor` to a specific `vimeo/psalm` version (the type declarations for the majority scenario are in the code, without any need for plugins)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+{
+    "version": 2,
+    "updates": [
+        {
+            "allow": [
+                {
+                    "dependency-type": "all"
+                }
+            ],
+            "directory": "/",
+            "package-ecosystem": "composer",
+            "schedule": {
+                "interval": "weekly"
+            },
+            "versioning-strategy": "widen"
+        },
+        {
+            "directory": "/",
+            "package-ecosystem": "github-actions",
+            "schedule": {
+                "interval": "weekly"
+            }
+        },
+    ]
+}
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -39,18 +39,7 @@ jobs:
           extensions: ${{ env.php-extensions }}
           ini-values: zend.assertions=1
 
-      - id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-
-      - name: Cache Composer dependencies
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: composer-${{ runner.os }}-${{ hashFiles('composer.json') }}
-          restore-keys: composer-${{ runner.os }}
-
-      - name: Install dependencies
-        run: composer update --prefer-dist --no-interaction
+      - uses: "ramsey/composer-install@v2"
 
       - name: Creating var directory
         run: mkdir -p var/cache
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "phpstan/phpstan": "^1.3",
         "phpstan/phpstan-strict-rules": "^1.0",
         "friendsofphp/php-cs-fixer": "^3.4",
-        "marcocesarato/php-conventional-changelog": "^1.12"
+        "marcocesarato/php-conventional-changelog": "^1.12",
+        "vimeo/psalm": "^4.18.1"
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9c97d188138f4d0c230b1b21b422ec51",
+    "content-hash": "013b98c33870edb3907acd6b581c06ae",
     "packages": [
         {
             "name": "doctrine/annotations",
@@ -80,32 +80,28 @@
         },
         {
             "name": "doctrine/lexer",
-            "version": "1.2.1",
+            "version": "1.2.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/lexer.git",
-                "reference": "e864bbf5904cb8f5bb334f99209b48018522f042"
+                "reference": "9c50f840f257bbb941e6f4a0e94ccf5db5c3f76c"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/lexer/zipball/e864bbf5904cb8f5bb334f99209b48018522f042",
-                "reference": "e864bbf5904cb8f5bb334f99209b48018522f042",
+                "url": "https://api.github.com/repos/doctrine/lexer/zipball/9c50f840f257bbb941e6f4a0e94ccf5db5c3f76c",
+                "reference": "9c50f840f257bbb941e6f4a0e94ccf5db5c3f76c",
                 "shasum": ""
             },
             "require": {
-                "php": "^7.2 || ^8.0"
+                "php": "^7.1 || ^8.0"
             },
             "require-dev": {
-                "doctrine/coding-standard": "^6.0",
-                "phpstan/phpstan": "^0.11.8",
-                "phpunit/phpunit": "^8.2"
+                "doctrine/coding-standard": "^9.0",
+                "phpstan/phpstan": "1.3",
+                "phpunit/phpunit": "^7.5 || ^8.5 || ^9.5",
+                "vimeo/psalm": "^4.11"
             },
             "type": "library",
-            "extra": {
-                "branch-alias": {
-                    "dev-master": "1.2.x-dev"
-                }
-            },
             "autoload": {
                 "psr-4": {
                     "Doctrine\\Common\\Lexer\\": "lib/Doctrine/Common/Lexer"
@@ -140,7 +136,7 @@
             ],
             "support": {
                 "issues": "https://github.com/doctrine/lexer/issues",
-                "source": "https://github.com/doctrine/lexer/tree/1.2.1"
+                "source": "https://github.com/doctrine/lexer/tree/1.2.2"
             },
             "funding": [
                 {
@@ -156,24 +152,24 @@
                     "type": "tidelift"
                 }
             ],
-            "time": "2020-05-25T17:44:05+00:00"
+            "time": "2022-01-12T08:27:12+00:00"
         },
         {
             "name": "psr/cache",
-            "version": "3.0.0",
+            "version": "1.0.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/php-fig/cache.git",
-                "reference": "aa5030cfa5405eccfdcb1083ce040c2cb8d253bf"
+                "reference": "d11b50ad223250cf17b86e38383413f5a6764bf8"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/php-fig/cache/zipball/aa5030cfa5405eccfdcb1083ce040c2cb8d253bf",
-                "reference": "aa5030cfa5405eccfdcb1083ce040c2cb8d253bf",
+                "url": "https://api.github.com/repos/php-fig/cache/zipball/d11b50ad223250cf17b86e38383413f5a6764bf8",
+                "reference": "d11b50ad223250cf17b86e38383413f5a6764bf8",
                 "shasum": ""
             },
             "require": {
-                "php": ">=8.0.0"
+                "php": ">=5.3.0"
             },
             "type": "library",
             "extra": {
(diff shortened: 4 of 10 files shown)


composer.json
```diff
@@ -27,7 +27,8 @@
         "phpstan/phpstan": "^1.3",
         "phpstan/phpstan-strict-rules": "^1.0",
         "friendsofphp/php-cs-fixer": "^3.4",
-        "marcocesarato/php-conventional-changelog": "^1.12"
+        "marcocesarato/php-conventional-changelog": "^1.12",
+        "vimeo/psalm": "^4.18.1"
```
Ocramius, line 31: This was added on purpose, since we are verifying type inference in `tests/StaticAnalysis`

tests/StaticAnalysis/can-infer-tree-mapper-results.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace CuyZ\Valinor\Tests\StaticAnalysis;
+
+use CuyZ\Valinor\Mapper\TreeMapper;
+use CuyZ\Valinor\Tests\StaticAnalysis\Stub\A;
+
+/** @param mixed $input */
+function mappingClassWillInferObjectOfSameType(TreeMapper $mapper, $input): A
```
Ocramius, line 11: Unsure if we need docs for this one - perhaps `README.md` in the directory :thinking:

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
+{
```
bendavies, line 1: json in a `.yml` file?

Ocramius: Yeah, makes it more manageable

bendavies: TIL. didn't know this was supported.

.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+                    "dependency-type": "all"
+                }
+            ],
+            "directory": "/",
+            "package-ecosystem": "composer",
+            "schedule": {
+                "interval": "weekly",
+                "time": "02:00"
+            },
+            "versioning-strategy": "widen"
```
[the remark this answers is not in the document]
romm: I'm not very familiar to Dependabot, if I understand correctly [the doc](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#versioning-strategy) you're suggesting to use `increase` instead?

Also, how does it work by default? A pull request will automatically be opened to bump to new versions?

.github/workflows/mutation.yml
```diff
@@ -39,18 +39,7 @@ jobs:
...
-      - name: Cache Composer dependencies
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: composer-${{ runner.os }}-${{ hashFiles('composer.json') }}
-          restore-keys: composer-${{ runner.os }}
-
-      - name: Install dependencies
-        run: composer update --prefer-dist --no-interaction
+      - uses: "ramsey/composer-install@v2"
```
romm, line 42: Noice! Didn't know that one.

src/Mapper/TreeMapper.php
```diff
@@ -7,11 +7,18 @@
 interface TreeMapper
 {
     /**
-     * @template T of object
+     * @psalm-template TObject of object
+     * @psalm-template TypeDefinition of string|class-string<TObject>
      *
-     * @param string|class-string<T> $signature
+     * @param TypeDefinition $signature
```
romm, line 13:
```suggestion
     * @param string|class-string<TObject> $signature
```
PhpStorm doesn't recognize the Psalm template and will not provide auto-completion if used. I'd prefer to keep this signature (unless you did it for a particular reason?). Also, the `@psalm-template TypeDefinition of string|class-string<TObject>` above should be removed.

composer.lock
```diff
@@ -203,31 +203,31 @@
...
-                "source": "https://github.com/php-fig/cache/tree/3.0.0"
+                "source": "https://github.com/php-fig/cache/tree/master"
             },
-            "time": "2021-02-03T23:26:27+00:00"
+            "time": "2016-08-06T20:24:11+00:00"
         },
         {
             "name": "psr/simple-cache",
-            "version": "2.0.0",
+            "version": "1.0.1",
```
romm, line 212: Looks like there are a lot of changes, possible you were using PHP < 8.1? I think it would be best to always lock to dependencies with PHP 8.1 (where all features are enabled). WDYT?

tests/StaticAnalysis/can-infer-tree-mapper-results.php
romm: Nah it's pretty clear IMO.

romm (commented):
Looking nice, thank you!

I'd like some explanation on the Dependabot thing, I often saw it on popular repos but I'm not sure how it works. 😅

Also, I try to keep using [conventional commits](https://www.conventionalcommits.org) for this repo (for consistency mainly). What would be your advice for this kind of PR where fixup commits were made? Should I squash it with a correct title/description when submitting the PR?

tests/StaticAnalysis/can-infer-tree-mapper-results.php
romm: Nah it's pretty clear IMO.

romm:
> Note: mutation tests fail because of no actual source diff in the patch.
> 
> Everything is done in:
> 
>     * `composer.json` + `composer.lock`
> 
>     * github actions manifests
> 
>     * dependabot manifest
> 
>     * docblocks

@Ocramius damn I don't get why it would work in #60 but not here. Any recommendation on how to fix this?

Ocramius:
> I'd like some explanation on the Dependabot thing, I often saw it on popular repos but I'm not sure how it works. sweat_smile

It will, once a week, look for new dependencies, `composer.json` and `composer.lock` accordingly (and also `.github/*` actions references), and send you PR for review for each dependency, so you don't need to chase changes in the ecosystem.

It will also rebase any PRs it currently has open, should `composer.json` and `composer.lock` change in between, so you don't need to do that work yourself.

 > Also, I try to keep using conventional commits for this repo (for consistency mainly).

I can squash out some commits, but no idea about what the actual commits here should be then? I keep them very much human-friendly, rather than "changed X", so if you want specific messages, please do tell.

 > What would be your advice for this kind of PR where fixup commits were made? Should I squash it with a correct title/description when submitting the PR?

Fixup commits will be squashed from my end, after review is through.

src/Mapper/TreeMapper.php
Ocramius: The point is giving PHPStan/Psalm more precise types, not to help PHPStorm.

If PHPStorm support is wished, I can provide a `.phpstorm.meta.php`, which does the auto-completion part: PHPStorm will eventually catch up with PHPStan/Psalm on this stuff.

composer.lock
Ocramius: I did a `composer update` after setting `{"config": {"platform": {"php": "7.4.99"}}}` on purpose, so that the resulting `composer.lock` can then be tested on `7.4`, `8.0` and `8.1`.

Ocramius:
Note: instead of a51f8c5 , I will rebase this patch post-review.

Merges of mainline into feature branches are really to be avoided (in pretty much every environment), and are a much bigger problem than conventional commits ;-)

src/Mapper/TreeMapper.php
Ocramius: I'll see if I can simplify the signature first, BTW :+1:

Ocramius:
> > Note: mutation tests fail because of no actual source diff in the patch.
> > Everything is done in:
> > ```
> > * `composer.json` + `composer.lock`
> > 
> > * github actions manifests
> > 
> > * dependabot manifest
> > 
> > * docblocks
> > ```
> 
> @Ocramius damn I don't get why it would work in #60 but not here. Any recommendation on how to fix this?

This is probably because I have a diff in `tests/`

src/Mapper/TreeMapper.php
romm: > The point is giving PHPStan/Psalm more precise types, not to help PHPStorm.

Sure, I got that, it's just that I believe we can have both PHPStan/Psalm support **and** PhpStorm autocompletion (very convenient for daily usage IMO).

The signature `string|class-string<TObject>` is understood by PhpStorm, and it looks like it would be ok for PHPStan/Psalm as well. Tell me if I'm wrong :D

composer.lock
romm: Oh ok I see, this is related to the changes to the dependencies matrix in the tests pipeline. 👍 

I'm not used to the usage of `{"config": {"platform": {"php": "7.4.99"}}}` — I guess you didn't commit that change because the dependencies upgrades will be managed by Dependabot, right?

Ocramius: Correct, I did:

 * manually add `{"config": {"platform": {"php": "7.4.99"}}}`
 * `composer update`
 * manually remove `{"config": {"platform": {"php": "7.4.99"}}}`
 * `composer update --lock`

That led to a PHP 7.4-compatible `composer.lock`

src/Mapper/TreeMapper.php
Ocramius: I just need to see if I can use the conditional here :+1:

.github/dependabot.yml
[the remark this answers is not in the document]
Ocramius: Explained in https://github.com/CuyZ/Valinor/pull/59#issuecomment-1010922020

src/Mapper/TreeMapper.php
Ocramius: 0bf2a13b8feac683276d8930da428f93c10246f8 simplifies this: solved

Ocramius:
> > I can squash out some commits, but no idea about what the actual commits here should be then? I keep them very much human-friendly, rather than "changed X", so if you want specific messages, please do tell.
> 
> The commits names are good, but what is missing is a correct prefix — see [conventional commits](https://www.conventionalcommits.org). This allows me to use [marcocesarato/php-conventional-changelog](https://github.com/marcocesarato/php-conventional-changelog) for the changelog automatic generation.

Rebased and added prefixes :muscle: 


> > Note: instead of [a51f8c5](https://github.com/CuyZ/Valinor/commit/a51f8c523bd07cc112ec278aaf83b5f70d1ce5cb) , I will rebase this patch post-review.
> > Merges of mainline into feature branches are really to be avoided (in pretty much every environment), and are a much bigger problem than conventional commits ;-)
> 
> Right, I usually use rebase strategies on my own branches, but I was not sure if that would fit on a PR managed by someone else. Learning basic OSS contribution rules here :D

Understandable: most people probably roll their eyes when I ask them for a rebase, but I do `git rebase` 10~20 times a day for my own stuff ;-)

> > This is probably because I have a diff in src/, but it doesn't involve any AST node changes, beside docblock
> 
> Ah! Right, must be that. Any idea how we could fix it, or avoid this case?

This is potentially a bug in `infection/infection` - I wouldn't worry too much about it, for now.

romm:
> Rebased and added prefixes 💪

Great! Not so important but FYI:
- For consistency the first letter of the subject is lowercase
- I force myself to have commit subjects that are less than 72 chars, this eases readability of the history — for instance GitHub will "break" subjects with more than 72 chars, this makes it harder to read:
  image

These are non-blocking from my POW but just wanted to share. 🙂 

> Understandable: most people probably roll their eyes when I ask them for a rebase, but I do `git rebase` 10~20 times a day for my own stuff ;-)

I also do, this is without doubt the best strategy to adopt, although rebasing on a contributor PR could also frighten the author if not used to it 😬 

> This is potentially a bug in `infection/infection` - I wouldn't worry too much about it, for now.

Alright, I'll see if I can report/fix it.

romm:
Just though I could easily reword the commits to fit the requirements I was writing about earlier — the sense doesn't change, it just fits the 72 chars threshold.  Hope you don't mind? I'm not sure if that can seem rude. 😬 

Anyway, thanks a lot for the contribution @Ocramius!

romm:
@Ocramius I'm not sure whether I should enable "Dependabot security updates" in the repository settings?

image

Ocramius:
Thanks @romm!

Enabling security updates shouldn't be necessary: weekly updates are enough for a library

PR #61, opened 2022-01-13
fix: correct regex that detects `@internal` or `@api` annotations

jdreesen:
The regex matched `@api` or `internal<whitespace>`.
Now it matches `@` followed by `api` or `internal` followed by `<whitespace>`.


romm:
Thank you, very much appreciated! ☺️

PR #63, opened 2022-01-24
fix: Attributes now combine correctly with foreign attributes

fred-jan:
Without this change I needed to define the Valinor attributes first, otherwise an undefined index error occurred (was using Doctrine annotations in our project).

### DIFF
--- a/src/Definition/AttributesContainer.php
+++ b/src/Definition/AttributesContainer.php
@@ -7,6 +7,7 @@ namespace CuyZ\Valinor\Definition;
 use Traversable;
 
 use function array_filter;
+use function array_values;
 use function count;
 
 /** @internal */
@@ -33,10 +34,10 @@ final class AttributesContainer implements Attributes
 
     public function ofType(string $className): iterable
     {
-        return array_filter(
+        return array_values(array_filter(
             $this->attributes,
             static fn (object $attribute): bool => $attribute instanceof $className
-        );
+        ));
     }
 
     public function count(): int
--- a/src/Mapper/Object/Factory/AttributeObjectBuilderFactory.php
+++ b/src/Mapper/Object/Factory/AttributeObjectBuilderFactory.php
@@ -33,6 +33,6 @@ final class AttributeObjectBuilderFactory implements ObjectBuilderFactory
             throw new TooManyObjectBuilderFactoryAttributes($class, $attributes);
         }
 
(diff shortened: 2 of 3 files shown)


romm:
Hi, thanks for the fix! Makes me think this issue is probably present at other places, I'll check it out.

Do you have time to add a unit test that would fail without your fix? If not I'll take care myself later on.

Also, I just checked out MyOnlineStore/common-domain#50, just so you know when #62 is merged there will be no need for the `StaticMethodConstructor` attribute anymore (it will even be deprecated). I'll try to make a new release today during lunchtime if it can help. 😉

romm:
Hi @fred-jan, I wanted to push some modification but it seems I don't have write permission on your branch; any chance you check the "Allow edits by maintainers" box? 🙂

romm:
Hello @fred-jan, just received your invitation to be a collaborator on your fork, thank you but that is unnecessary: you should be able to give me the rights just here in the pull request, see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Normally there should be a `Allow edits from maintainers` checkbox there. 🙂

fred-jan:
>Hi @fred-jan, I wanted to push some modification but it seems I don't have write permission on your branch; any chance you check the "Allow edits by maintainers" box? slightly_smiling_face

HI @romm, I was looking for that option, and also read that specific article, but for some reason I don't have that option available to me. I couldn't find a way to enable that option, so I gave you temporary write permissions instead.

romm:
Really strange indeed, well at least now it worked. 🙂 

I preferred to fix the issue right at the source (see 5afd1ce), what do you think?

fred-jan:
> Right, must be that! That's a pity though.

That seems to be the issue really, judging by this post: https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847

> I preferred to fix the issue right at the source (see 5afd1ce), what do you think?

Even better! :ok_hand: I also considered to re-index the attributes while passing them to the constructor of `ClassDefinition`, but I guess by doing that you'd lose that information might you ever need it (the relative position of an attribute in respect to other attributes).

romm:
Yeah I think this is a better place to do it, furthermore this is a filter so when using this method the position would not be true anymore anyway.

Thanks for the contribution!

Issue #65, opened 2022-01-24
Problems with $cacheDir attribute


marek-mikula:
Hello,

I was having some problems with permissions on my temp folder on my Mac so I decided to provide my own custom cache folder for compiled files.

But I ran into an issue. The cached string is firstly put into temp folder and then 'renamed' (moved into my specified cache folder') so I can't get passed the error with my permissions.

Is there any specific reason why to do this way? Why not just directly put the cached string into the specified cache folder?

Thank you in advance and thank you for this package, it's really cool :)

https://github.com/CuyZ/Valinor/blob/e2800339411b68a24eebc266377184c76ef2701e/src/Cache/Compiled/CompiledPhpFileCache.php#L85

romm:
Hi @marek-mikula, indeed I never ran into this issue myself but seems like some change should be done to prevent this usecase to happen.

Would you mind opening a PR with a fix that runs on your application? It looks like the test suite would not need any modification.

marek-mikula:
Hello @romm, thank you for your immediate response. I will open PR today after work.

romm:
That would be very much appreciated, thanks in advance!

PR #67, opened 2022-01-27
feat: cached file is saved directly without need of moving the file

marek-mikula:
I replaced the steps of saving and moving the file with only one step of saving the file directly using file_put_contents().


romm:
Hi @marek-mikula, thanks for the contribution! 🙂 

However the rename was done on purpose, to avoid concurrent writing to the cache file, so this must stay; the temporary file can still be created in the configured cache directory though.

romm:
Just done some changes that better fit how the system was supposed to work. Thanks for the initial work @marek-mikula!

closes #65

marek-mikula:
@romm Hello, I am really sorry, I actually totally forgot about this. Thank you for your work! I really appreciate it!

PR #69, opened 2022-01-28
feat: introduce a camel case key modifier as a source

Mopolo:


romm:
Just done some changes, WDYT?

src/Mapper/Source/Modifier/CamelCaseKeys.php
```diff
@@ -0,0 +1,63 @@
...
+    private function replace(iterable $source): array
+    {
+        $result = [];
+
+        foreach ($source as $key => $value) {
+            if (is_iterable($value)) {
+                $value = $this->replace($value);
+            }
+
+            $camelCaseKey = $this->camelCaseKeys($key);
```
romm, line 42: I was wondering: maybe we should delete the original key, it would probably reduce the memory consumption.

Mopolo: You mean doing a `unset($source[$key]);` I suppose?

romm: Yes, next to `$result[$camelCaseKey] = $value;`. A new unit (not integration) test should be provided though.

Mopolo: If `$source` is an instance of `\ArrayAccess` the `unset` will change the object outside the modifier class.

Is that a good idea? Maybe we should only do it if `$source` is an `array` to avoid side effects?

romm: 👍

PR #71, opened 2022-02-15
Ensure mixed types remain valid

mtouellette:
i.e. not mixed|null


romm:
Thank you, that's a good catch!

Would you mind adding a unit test case in `ReflectionTest`?

mtouellette:
Sure.  I have added a test.

romm:
Very appreciated, thanks!

PR #82, opened 2022-03-01
Fix camel key modifier when used with non associative array

magnetik:
When using the camel key modifier with a source like:

```
{
  "foo_bar": "test",
  "test": [
     { "foo": "bar" },
     { "foo": "bar" }
   ]
}
```

The camel key modifier throws an exception. 

This fixes the behaviour


romm:
Thank you guys!

PR #83, opened 2022-03-02
feat: introduce a source builder

Mopolo:
It helps building the source object like this:

```php
$source = \CuyZ\Valinor\Mapper\Source\Source::json($json)
    ->map(['foo-key' => 'bar-key'])
    ->camelCaseKeys();
```


src/Mapper/Source/Source.php
```diff
@@ -0,0 +1,63 @@
...
+    private IteratorAggregate $delegate;
+
+    /**
+     * @param IteratorAggregate<mixed> $delegate
+     */
+    private function __construct(IteratorAggregate $delegate)
+    {
+        $this->delegate = $delegate;
+    }
+
```
romm, line 30: I think we should also have named constructors for `array` and/or `iterable`. WDYT?

README.md
```diff
@@ -408,6 +408,38 @@ $source = new \CuyZ\Valinor\Mapper\Source\Modifier\PathMapping([
     ->map(Country::class, $source);
 ```
 
+#### Source builder
```
romm, line 411: The builder should probably be the new default entry point IMO (?), so the documentation above should be modified I think.

romm (changes requested):

README.md
```diff
@@ -385,15 +389,18 @@ final class City
...
         ['label' => 'Ankh Morpork'],
         ['label' => 'Minas Tirith'],
     ],
-], [
-    'towns' => 'cities',
-    'towns.*.label' => 'name',
-]);
+];
+
+$source = \CuyZ\Valinor\Mapper\Source\Source::array($input)
```
romm, line 399: Let's put the input directly in the method call? WDYT?

README.md
```diff
@@ -348,14 +349,17 @@ final class SomeClass
...
     'some_value' => 'foo',
     // …or…
     'some-value' => 'foo',
     // …or…
     'some value' => 'foo',
     // …will be replaced by `['someValue' => 'foo']`
-]);
+];
+
+$source = \CuyZ\Valinor\Mapper\Source\Source::array($input)
```
romm, line 361: Let's put the input directly in the method call? WDYT?

romm (approved):

Issue #44, opened 2021-12-19
Strengthen tests for datetime mapping


romm:
As stated by @keichinger in https://github.com/CuyZ/Valinor/pull/40#discussion_r771919725, the test-cases below can lead to false-positive, mainly because every case is mapping to the exact same datetime.

https://github.com/CuyZ/Valinor/blob/f12ef39e2f92667d97e84e632f8a08e617252d2a/tests/Integration/Mapping/Type/DateTimeMappingTest.php#L18-L60

Changing each case to a different datetime everytime would strengthen the tests and ensure the expected behaviour is correct.

romm:
Fixed in #84, thanks @Lucian-Olariu!

PR #84, opened 2022-03-02
Strengthen datetime mapping tests

Lucian-Olariu:
Addresses https://github.com/CuyZ/Valinor/issues/44

### DIFF
--- a/tests/Integration/Mapping/Object/DateTimeMappingTest.php
+++ b/tests/Integration/Mapping/Object/DateTimeMappingTest.php
@@ -11,23 +11,23 @@ use DateTime;
 use DateTimeImmutable;
 use DateTimeInterface;
 
-use function mt_rand;
+use function random_int;
 
 final class DateTimeMappingTest extends IntegrationTest
 {
     public function test_datetime_properties_are_converted_properly(): void
     {
-        $dateTimeInterfaceValue = $this->get_random_timestamp();
-        $dateTimeImmutableValue = $this->get_random_timestamp();
-        $dateTimeFromTimestampValue = $this->get_random_timestamp();
-        $dateTimeFromTimestampWithFormatValue = (new DateTime())->setTimestamp($this->get_random_timestamp())->format(DATE_ATOM);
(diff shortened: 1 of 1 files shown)


tests/Integration/Mapping/Object/DateTimeMappingTest.php
```diff
@@ -11,24 +11,34 @@
 use DateTimeImmutable;
 use DateTimeInterface;
 
+use function mt_rand;
```
Ocramius, line 14: `mt_rand` is long gone - let's use `random_int()` instead

tests/Integration/Mapping/Object/DateTimeMappingTest.php
```diff
@@ -108,6 +118,11 @@ public function test_invalid_array_source_throws_exception(): void
             self::assertSame('1607027306', $error->code());
         }
     }
+
+    private function get_random_timestamp(): int
```
Ocramius, line 122: Perhaps rather than `get`, `makeRandomTimestamp()`?

tests/Integration/Mapping/Object/DateTimeMappingTest.php
```diff
@@ -108,6 +118,11 @@ public function test_invalid_array_source_throws_exception(): void
...
+    {
+        return mt_rand(1, time());
```
Ocramius, line 124: Not sure whether this needs to be restricted to `time()` as an upper boundary: IMO worth testing also in the far future (year 3000 or so)

Ocramius (changes requested):

tests/Integration/Mapping/Object/DateTimeMappingTest.php
Lucian-Olariu: Using `random_int()`

romm:
Hi @Lucian-Olariu, very good contribution, I just cleaned up a bit to ease readability. Thanks a lot! 🙂

Lucian-Olariu:
Yes wanted to do that but thought more tests will come later and  that is why I declared them in separate variables :+1:

Issue #85, opened 2022-03-03
Named constructor that received input as array


magnetik:
Hi,

I've an object which deserialization can only be done in a constructor that receive all sub elements as an array.

Something like

```
{
  "foo": {
     "test": true,
     "other": "yes", 
     "something_else": false
   }
}
```

That is handled by :

```
class Foo {
  public static function fromArray(array $data) {
    $result = new self();
    $result->somethingSuperStrange = $data['test'] ?? $data['something_else']
    return $result;
  }
}
```

Don't mind the something super strange as this is not my real use case, I really need to handle some weird data.

Can something like this be done with valinor ?

romm:
Hi @magnetik, the goal of the library is to help with the mapping of the data, so I guess the constructor would be better with something like below, although I'm not sure how useful this kind of usecase can be.

```php
class Foo {
    public static function fromArray(string $other, bool $somethingElse, bool $test) {
        $result = new self();
        $result->somethingSuperStrange = $test || $somethingElse;

        return $result;
    }
}
```

Note that the parameters can also be optional, depending on the needs, in which case they don't need to be in the source:

```php
class Foo {
    public static function fromArray(string $other, bool $somethingElse = false, bool $test = false) {
        $result = new self();
        $result->somethingSuperStrange = $test || $somethingElse;

        return $result;
    }
}
```

I'm not sure I'm answering your question, please tell me if I can be of a better help. 🙂

magnetik:
Thanks for your answer.

I've tried to simplify my example but it was not relevant ^^

The real thing I want to tackle is an object containing properties with any name possible.
For instance I can have :
```
{ 
  "ActivateThingy":  true,
  "SomethingElse": 41,
  "DifferentConfig": "A string"
}
```

and I map them in an object implementing `ArrayAccess` . 

Not sure what would be the right way to map it using valinor?

romm:
Sorry I'm still struggling to understand properly. 😬 

You do have to specify all properties/arguments on which the source will be mapped to; what do you mean by "any name possible"? Is this some kind of "dynamic property naming" or something? 

May I ask the real usage/application of your example? Maybe with a bigger picture I could understand more easily.

magnetik:
I'm failing to explain it clearly, no problem :D 

It's dynamic properties indeed !

My application can receive an object with any arbitrary key.
(I'm forced to use this kind of input by the application using my API)

romm:
I see, but then how are you supposed to know which type to apply on properties?

[...]

PR #87, opened 2022-03-08
Implement an hackish ArrayAccess mapper

magnetik:
A quick and dirty object mapper regarding.

Not sure if you want it in Valinor, but it does the job for what I wanted in #85


romm:
Closing, see https://github.com/CuyZ/Valinor/issues/85#issuecomment-1061645080.

PR #102, opened 2022-04-01
fix: only call matching callback if value is accepted

Mopolo:


romm:
I've just done some modification, checking if types were matching was actually not necessary when also checking if the value was accepted.

Otherwise, thanks for the fix!

Issue #115, opened 2022-05-02
Pre-generation of signature caches


boesing:
Hey there,

I recently started to use this library.
I was thinking of actually writing a CLI command for CI pipeline which searches for the `TreeMapper#map` usage and parses the 1st argument (FQCN) from it.

I do want to pass this FQCN to the `TypeParser#parse` method so that the `cacheDirectory` entry is being created.

What I found out is, that there is no way to actually get an instance of the `TypeParser` and I do really would love to avoid instantiating it by myself.

Do you have an idea and/or suggestion on how to retrieve the `TypeParser` instance with the settings I do pass to the `MapperBuilder`?

Would love to get some help here so we can pre-generate the cache entries on the filesystem during build-time.

Thanks in advance 🤙🏼

romm:
Hi @boesing, I did not give access to the container instance on purpose to avoid giving access to everything (which would result in more maintenance burden).

However, this is a good suggestion, and a part of this feature could even be part of the public API provided by the library, maybe something like `MapperBuilder::warmup(SomeClass::class, SomeOtherClass::class)`. I'd be happy to help/review if you were to add this to the core of the library. 🙂

romm:
Closed with #116.

PR #116, opened 2022-05-03
Feature: Signature cache warmup via `MapperBuilder#warmup`

boesing:
As requested in #115, it would be awesome to have the ability to warmup known signatures (e.g. during build time).

One [suggestion](https://github.com/CuyZ/Valinor/issues/115#issuecomment-1115824858) was to provide a dedicated method for this in the `MapperBuilder` (which is also used to specify the cache directory).

So this Pull-Request does provide that `MapperBuilder#warmup` method and consumes at least 1 but also multiple signature(s) which will be parsed and pre-cached to the underlying filesystem cache.


romm:
Hi @boesing, very good work overall!

I think one interesting feature would be to loop on every type that was declared inside classes to recursively warmup every class.

For instance:

```php
final class Foo
{
    public Bar $bar;
}

final class Bar
{
    public string $value;
}
```

If the `warmup` method is called for `Foo`, then the `Bar` class should be warmed up as well.

As we need to handle complex types (union types containing objects, or array shapes containing objects), I will add a helper for this and keep you updated here. 🙂

boesing:
I tried to wrap my head around this yesterday evening.
I wanted to ensure that we do not double-warmup the same class-name and thus the code became quite - complex...
Maybe we just don't care about double-warmup (as the memory cache might kick-in anyways) and thus we can just pass every `ClassType` to the repository.

Will see if I give this another try in the next days.

romm:
I think it would not be an issue to warmup the same class multiple times, but I also think it would be really easy to prevent that from happening — keeping a list of already loaded classes up to date should be enough IMO.

The code that does the job should also be moved in a specific service, to prevent `MapperBuilder` from becoming too huge. 🙂 

If you need any advice or help please don't hesitate!

boesing:
I guess I got a working prototype. 
Would love to get some feedback.

romm:
Hi @boesing, thanks to #132 the implementation could be changed in a cleaner way, can you take a look if you're good with it?

src/MapperBuilder.php
```diff
@@ -250,6 +250,17 @@ public function withCacheDir(string $cacheDir): self
         return $this->withCache(new FileSystemCache($cacheDir));
     }
 
+    /**
+     * Warms up the injected cache implementation with the provided class names.
+     *
+     * By passing a class which contains recursive objects, every nested object
+     * will be cached as well.
+     */
+    public function warmup(string ...$signatures): void
```
romm, line 259: Note: the signature changed, compared to what was done originally:

https://github.com/CuyZ/Valinor/blob/b11eaf0bcd247f1d584fbee7b42b5d0b05b09a9b/src/MapperBuilder.php#L282-L295

Two things:

1. I changed the `class-string` signature, because one could want to warmup a signature containing generics, for instance `SomeClass<int>` and `SomeClass<string>` should both be warmed up (as they are handled in separate cache files — unfortunately).
2. Although it looked like a good idea to force at least one argument, I think that this can lead to an undesired side effect; for instance when dynamically building a list of classes to be warmed up, then passing it to the method — if there was actually no class to warm up (in a fresh installed framework for instance), that would lead to a fatal error.

boesing: Works for me.

boesing:
LGTM, the changes are still reflecting the initial feature request. 👍🏼 
Thanks for taking over here!

romm:
Thank you for the original work, that was a great contribution!

PR #129, opened 2022-05-11
[Rector] Apply Rector for QA

samsonasik:
Apply Rector for QA in Github workflow, start with `LevelSetList::UP_TO_PHP_74`. 

Use pinned version 0.12.23 on purpose as it need to sync phpstan and nikic/php-parser, ensure only run when working ok.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,7 +12,6 @@
 /phpstan.neon.dist export-ignore
 /psalm.xml export-ignore
 /phpunit.xml.dist export-ignore
+/rector.php export-ignore
 
 *.stub linguist-language=PHP
-
-/rector.php
--- a/.gitattributes
+++ b/.gitattributes
@@ -14,3 +14,5 @@
 /phpunit.xml.dist export-ignore
 
 *.stub linguist-language=PHP
+
+/rector.php
--- a/.github/workflows/quality-assurance.yml
+++ b/.github/workflows/quality-assurance.yml
@@ -47,14 +47,14 @@ jobs:
 
       - uses: "ramsey/composer-install@v2"
 
-      - name: Running PHP Coding Standards Fixer
-        run: php vendor/bin/php-cs-fixer fix --dry-run
-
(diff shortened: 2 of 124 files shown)


samsonasik:
All green 🎉

romm:
Hi @samsonasik thanks for the contribution, although I'm not sure why Rector would be useful here? The library already uses [friendsofphp/php-cs-fixer](https://github.com/CuyZ/Valinor/blob/a6c60eb60b8ffcc8cdbbfaecc9e0884ac2aaf512/composer.json#L29) to check for style issues, what more does it bring to the workflow?

My first thought about Rector would be to use it to easily convert the codebase when bumping the minimum PHP requirement (which is great but not really the case here), maybe I'm missing something.

samsonasik:
it can cover new syntax update for this initial PR, eg: arrow function, ref https://github.com/CuyZ/Valinor/pull/129/files#diff-fbf35dae269032cbbaa34396944039b84c4670c9f6552a9f6cdf27561efc7b0b

it can detect dead code too, etc, that I may can provide another PR for it with dead code setlist.

romm:
Alright, after all there are never enough static analysis. 😁 

One thing though: all these changes for the timestamp values (in exceptions) are not needed IMO, can you update the settings so that it doesn't apply on it?

Also, running Rector in `dry-run` mode should be part of the [check](https://github.com/CuyZ/Valinor/blob/a6c60eb60b8ffcc8cdbbfaecc9e0884ac2aaf512/composer.json#L46-L51) Composer script ; running it to actually fix the files should be part of the [fix](https://github.com/CuyZ/Valinor/blob/a6c60eb60b8ffcc8cdbbfaecc9e0884ac2aaf512/composer.json#L52-L54) script.

samsonasik:
@romm implemented 👍

romm:
Hi @samsonasik, just did some adjustments.

Thanks for the contribution!

PR #130, opened 2022-05-16
Feature: setting to disable cache source validation

boesing:
In #116, a new feature `MapperBuilder#warmup` will be introduced, which allows projects to pre-generate cached class definitions within their CI pipeline.

In some projects, this might lead to the problem, that the pipeline is being executed on a different server and thus probably in a different location (path).

The current cache implementation provides an ability to verify if the cache is still "up2date". This is done by comparing the current `filemtime` of the cache-source file in combination with the timepoint of that moment, when the cache file was created/generated.

I'd like to propose a setting to disable that cache validation as (at least in our case) the source-code does not change when its deployed to production. Having some kind of `filemtime` check in non-production systems make absolutely sense and thus this validation should be enabled by default (as it was before).

Happy to get some feedback here.

### DIFF
--- a/src/Definition/Repository/Cache/Compiler/ClassDefinitionCompiler.php
+++ b/src/Definition/Repository/Cache/Compiler/ClassDefinitionCompiler.php
@@ -28,13 +28,16 @@ final class ClassDefinitionCompiler implements CacheCompiler, CacheValidationCom
 
     private PropertyDefinitionCompiler $propertyCompiler;
 
-    public function __construct()
+    private bool $validateCacheSource;
+
+    public function __construct(bool $validateCacheSource)
     {
         $this->typeCompiler = new TypeCompiler();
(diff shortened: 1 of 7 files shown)


romm:
Hi @boesing, thanks a lot for the contribution.

As stated in https://github.com/CuyZ/Valinor/issues/131#issuecomment-1133629748, I already thought about another way of dealing with caches and invalidation in general. I'll soon be submitting a patch to change the behavior, that will make this PR obsolete.

In the meantime I'll keep it open, and when the patch comes out I'd be pleased to have your feedback. 🙂

romm:
Hi @boesing, just submitted #132, could you take a look and give some feedback if needed?

boesing:
Ah, sorry! I had a quick glance on what you changed in #132 and that totally suids my needs.

Closing this in favor of #132

PR #135, opened 2022-05-24
support @var annotion for property promotion parameter

DavidBadura:
Hello, I encountered a problem while trying out the interesting library. I use a lot of dtos with property promotions, my typing was not recognized there and eg. `non-empty-string` was ignored (missing mapping error message).

```php
final class PublishMessage
{
    public function __construct(
        /** @var non-empty-string */
        public readonly string $text
    ) {
    }
}
```

I dug deeper and saw what the problem was. I noticed that in the case of property promotion, the @var annotation is ignored when the constructor is used.

I am of the opinion that with promoted properties the parameters should be treated exactly as properties and thus @var annotation are allowed. So do other tools like phpstan/psalm etc.

I created a first fix, which I would be happy to improve with your feedback.

### DIFF
--- a/src/Utility/Reflection/Reflection.php
+++ b/src/Utility/Reflection/Reflection.php
@@ -120,6 +120,7 @@ final class Reflection
 
         if (method_exists($reflection, 'isPromoted') && $reflection->isPromoted()) {
             $type = self::parseDocBlock(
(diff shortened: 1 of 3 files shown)


romm:
Perfect, thank you!

That bug was already reported by some people, so that's even better!

PR #136, opened 2022-05-24
Support mapping from json encoded mapping result

mtouellette:
useful in api integration tests

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -30,4 +30,4 @@ parameters:
 
     stubFiles:
         - qa/PHPStan/Stubs/Psr/SimpleCache/CacheInterface.stub
-    tmpDir: var/cache/phpstan
+    tmpDir: c:\Temp\cache\phpstan
--- a/src/Mapper/Object/DateTimeImmutableJsonSerializable.php
+++ b/src/Mapper/Object/DateTimeImmutableJsonSerializable.php
@@ -23,7 +23,7 @@ final class DateTimeImmutableJsonSerializable extends \DateTimeImmutable impleme
      * @return DateTimeImmutableJsonSerializable|false
      * @throws Exception
      */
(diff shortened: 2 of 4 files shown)


romm:
Hi @mtouellette, thank you for the contribution. 

Although I understand your needs, this pull request cannot be merged as it is: the library wont support custom oriented implementations of `DateTimeInterface`, even if this can be useful in your case.

Still, I think you can manage to make it working on your side, can you try this in your application:

```php
try {
    $result = (new \CuyZ\Valinor\MapperBuilder())
        ->infer(DateTimeInterface::class, fn() => SomeJsonSerializableDateTime::class)
        ->mapper()
        ->map(SomeClass::class, [/* … */]);
} catch (MappingError $error) {
    // …
}
```

But your proposition reminded me that the `infer` method actually doesn't support class inheritance, which is a pity. The code below **doesn't work** for now, but I will certainly add this feature at some point.

```php
try {
    $result = (new \CuyZ\Valinor\MapperBuilder())
        ->infer(DateTime::class, fn() => SomeClassThatInheritsDateTime::class)
        ->mapper()
        ->map(SomeClass::class, [/* … */]);
} catch (MappingError $error) {
    // …
}
```

mtouellette:
It looks like it might, though I haven't tried yet.  I'm dealing with a lot of auto generated code from an OpenApi specification and the output of the data transfer objects uses DateTime rather than DateTimeInterface.  Will infer work with concrete classes as well as interfaces?

romm:
Ok, thanks for the info.

That's what I wrote above, unfortunately inferring wont work with classes inheritance for the moment; this is clearly a missing feature that I will be adding to the library at some point: check for next releases!

mtouellette:
Ah, I see.  I'll certainly keep an eye out for updates.  Thanks again!

PR #138, opened 2022-05-25
DateTimeObjectBuilder: support dates without time

Slamdunk:
Hi, here I propose the support for dates without time on `DateTimeObjectBuilder`.

### DIFF
--- a/src/Mapper/Object/DateTimeObjectBuilder.php
+++ b/src/Mapper/Object/DateTimeObjectBuilder.php
@@ -27,6 +27,7 @@ final class DateTimeObjectBuilder implements ObjectBuilder
 {
     public const DATE_MYSQL = 'Y-m-d H:i:s';
     public const DATE_PGSQL = 'Y-m-d H:i:s.u';
+    public const DATE_WITHOUT_TIME = '!Y-m-d';
 
     /** @var class-string<DateTime|DateTimeImmutable> */
     private string $className;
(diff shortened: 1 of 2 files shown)


src/Mapper/Object/DateTimeObjectBuilder.php
```diff
@@ -27,6 +27,7 @@ final class DateTimeObjectBuilder implements ObjectBuilder
 {
     public const DATE_MYSQL = 'Y-m-d H:i:s';
     public const DATE_PGSQL = 'Y-m-d H:i:s.u';
+    public const DATE_WITHOUT_TIME = '!Y-m-d';
```
Slamdunk, line 30: The `!` exclamation mark is intended to **not** get hours, minutes and seconds from current time, see for example https://github.com/doctrine/dbal/blob/4c1926e26ef11a3119593ad81490d4dfe22618d8/src/Types/DateImmutableType.php#L50

tests/Integration/Mapping/Object/DateTimeMappingTest.php
```diff
@@ -63,6 +66,7 @@ public function test_datetime_properties_are_converted_properly(): void
         self::assertEquals(DateTimeImmutable::createFromFormat($dateTimeFromArray['format'], $dateTimeFromArray['datetime']), $result->dateTimeFromArray);
         self::assertEquals(DateTimeImmutable::createFromFormat(DateTimeObjectBuilder::DATE_MYSQL, $mysqlDate), $result->mysqlDate);
         self::assertEquals(DateTimeImmutable::createFromFormat(DateTimeObjectBuilder::DATE_PGSQL, $pgsqlDate), $result->pgsqlDate);
+        self::assertSame($sqlDateNotTime . ' 00:00:00', $result->sqlDateNotTime->format(DateTimeObjectBuilder::DATE_MYSQL));
```
Slamdunk, line 69: `' 00:00:00'` test part is intended to cover the `!` exclamation mark on format

romm:
LGTM

I'll definitely think about a better way to give the user more flexibility on what format(s) can be used to map dates; this is a pity that you had to open a PR for this — but in the meantime this is much appreciated, thanks!

PR #139, opened 2022-05-26
MappingError: expose the first error in Exception message

Slamdunk:
Hi, the [message customization](https://github.com/CuyZ/Valinor#message-customization) is great, but a first feedback without it would be very useful to speed up debugging.

Before and after this PR:

```diff
-CuyZ\Valinor\Mapper\MappingError: Could not map type `array{foo: string, bar: int}` with the given source.
+CuyZ\Valinor\Mapper\MappingError: Could not map type `array{foo: string, bar: int}` with the given source. First error is:
+Path: foo
+Given value: array (empty)
+Error: Cannot be empty and must be filled with a value matching type `string`.
```

### DIFF
--- a/tests/Integration/Mapping/MappingErrorTest.php
+++ b/tests/Integration/Mapping/MappingErrorTest.php
@@ -10,22 +10,42 @@ use CuyZ\Valinor\Tests\Integration\IntegrationTest;
 
 final class MappingErrorTest extends IntegrationTest
 {
-    public function test_first_error_is_reported_in_exception_message(): void
+    public function test_mapping_error_code_is_correct(): void
     {
         try {
-            (new MapperBuilder())
-                ->mapper()
-                ->map(
-                    'array{foo: string, bar: int}',
-                    ['foo' => [], 'bar' => []]
-                );
+            (new MapperBuilder())->mapper()->map('string', ['foo']);
+
             self::fail();
         } catch (MappingError $exception) {
-            self::assertStringContainsString('Path: foo', $exception->getMessage());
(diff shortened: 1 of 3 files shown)


romm:
Hi @Slamdunk, that was a very good idea. I took some time to think about it and reworked a bit around your initial contribution.

Hope you'll like it!

tests/Integration/Mapping/MappingErrorTest.php
```diff
@@ -0,0 +1,51 @@
...
+final class MappingErrorTest extends IntegrationTest
+{
+    public function test_mapping_error_code_is_correct(): void
+    {
+        try {
+            (new MapperBuilder())->mapper()->map('string', ['foo']);
+
+            self::fail();
+        } catch (MappingError $exception) {
+            self::assertSame(1617193185, $exception->getCode());
```
Slamdunk, line 20: `try-catch` statements inside tests may be misleading: did the author want to test the Exception, or just protect the test from and edge case?
This one may be easy to read, but in the long run it becomes harder: I suggest to leverage the built-in expectations next time: https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#testing-exceptions

Slamdunk (commented):
Thank you @romm, awesome improvements over my first implementation :muscle:

tests/Integration/Mapping/MappingErrorTest.php
romm: True, I actually wonder why I didn't go this way. Gonna make the change now!

PR #165, opened 2022-06-28
Fixing example of typed array mapping

cajax:
Hi!

The given example with `@var` doesn't map answers items to `Answer` class. They remain plain arrays. 
Instead of this I tried with `@param` and it did work well.

TBH I didn't dig too much in code. Maybe this example was written before something changed in Valinors core.

### DIFF
--- a/docs/pages/getting-started.md
+++ b/docs/pages/getting-started.md
@@ -43,11 +43,13 @@ A schema representing the needed structure must be provided, using classes.
 ```php
 final class Thread
 {
+    /**
+     * @param Answer[] $answers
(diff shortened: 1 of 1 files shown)


romm:
Hi @cajax, thanks for the contribution, however this issue should have been fixed with #135. What version are you using? This should be fixed since version [0.10.0](https://github.com/CuyZ/Valinor/releases/tag/0.10.0).

romm:
Just checked again against current `master` and the code in the documentation works fine. Thanks for the report though. 🙂

PR #186, opened 2022-07-24
Extract promoted property/param attribute based on the attribute target flags

rguennichi:
First of all, I wanted to thank everyone who has contributed to this amazing library! Good job guys!

I created this PR after dealing with some issues while trying to use `symfony/validator` with Valinor:

```php
use App\Request\ParamConverter\RequestQueryInterface;
use Symfony\Component\Validator\Constraints\Range;

class TestQuery implements RequestQueryInterface
{
    public function __construct(
        #[Range(min: 10)]
        public readonly int $page,
    ) {
    }
}
```

After calling `map()` method I received this exception:

```
Attribute "Symfony\Component\Validator\Constraints\Range" cannot target parameter (allowed targets: method, property)
```

So basically, all Symfony's constraint attributes are declared with these flags:
```php
#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
```

And based on [PHP RFC: Constructor Property Promotion #Attributes](https://wiki.php.net/rfc/constructor_promotion#attributes) promoted properties with attributes will apply to both **property** AND **parameter** reflections, which I hope it'll get improved in the upcoming PHP versions. So in order to fix this issue, I though it should be useful to prepare quick workaround by excluding attributes with errors while trying to create them via `$attribute->newInstance()` call.

### DIFF
--- a/src/Definition/NativeAttributes.php
+++ b/src/Definition/NativeAttributes.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CuyZ\Valinor\Definition;
 
 use CuyZ\Valinor\Definition\Exception\InvalidReflectionParameter;
+use Error;
 use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionMethod;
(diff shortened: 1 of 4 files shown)


rguennichi:
Another clean alternative for this maybe to have Valinor's super class attribute extended by all internal attributes, and then we can filter attributes based on this parent attribute, and that way we keep third party's attributes extracted/handled normally by their extractors/handlers.

Example:

Parent attribute:
```php  
namespace CuyZ\Valinor\Attribute;

abstract class Attribute
{
}
```

Current `Identifier` attribute will extend this abstract `Attribute` class:
```php
#[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_PARAMETER)]
final class Identifier extends Attribute implements ShellVisitor
{
    public function visit(Shell $shell): Shell
    {
        return $shell->withValue($shell->parent()->name());
    }
}
```

Now on `NativeAttributes` we could filter Valinor's attributes by:
```php
/**
 * @return array<ReflectionAttribute<object>>
 */
private function attributes(Reflector $reflection): array
{
    foreach ([ReflectionClass::class, ReflectionProperty::class, ReflectionMethod::class, ReflectionParameter::class] as $classname) {
        if (!$reflection instanceof $classname) {
            continue;
        }

        return $reflection->getAttributes(Attribute::class, ReflectionAttribute::IS_INSTANCEOF);
    }

    throw new InvalidReflectionParameter($reflection);
}
```

romm:
Hi @rguennichi, thanks for the excellent explanation and the contribution. Looks legit to me.

I'll give it some thoughts and try to review and merge it in the upcoming days, then publish a new release of the library.

romm:
Thanks @rguennichi!

PR #193, opened 2022-07-31
release: version 0.13.0

romm:


szepeviktor:
Hello @romm!
I've heard people stay away from Valinor because it is changing too fast.
What do you think?

romm:
Hi @szepeviktor, thanks for posting.

I can relate, that is mainly because the library is in a beta state, so I take all opportunities to fix bad designs while I can do it easily. Also, [two security issues](https://github.com/CuyZ/Valinor/security/advisories?state=published) were discovered and needed breaking changes to be fixed; I'm not happy about it but I think one can understand this kind of change.

I plan to release a first stable version as soon as possible (I'm hoping end of the year); when this happens, breaking changes will be impossible without a new major release, so it will be easier for users to follow changes. I believe that the library is becoming very mature and less prone to major changes anyway.

People that need stability and don't want to handle the hassle of upgrading minor versions with potential BC should wait for the first stable version, although I still hope that some will remain and help me reaching the 1.0 milestone. 😉 

Cheers.

PR #194, opened 2022-08-01
Introduce configurable date formats for `DateTimeObjectBuilder`

boesing:
Hey there,

since mapping of external API responses should be as performant as possible, I'd really love to configure the formats the `DateTimeObjectBuilder` should actually try to apply.

In our API specs, we do always have either RFC-3339 or `Y-m-d` (only date). Depending on other APIs, maybe other formats are being preferred. To avoid multiple `DateTimeImmutable::createFromFormat` executions with unnecessary formats which are not configured either, I was thinking about adding a way to actually configure these formats globally.

This is actually implemented in a BC compatible way, so unless there are specific formats configured, the "default" formats (as before) are being used.

I have no idea where the best place to actually document this feature would be. I'd say in the README but since there are not all features document, I was unable to find a slot where I could just append this new setting.

Happy to get some feedback.
Until then, thanks for all the work here - this is one of those libraries which had the most impact in my development experience.

### DIFF
--- a/src/Library/Container.php
+++ b/src/Library/Container.php
@@ -153,7 +153,11 @@ final class Container
 
                 $factory = new ReflectionObjectBuilderFactory();
                 $factory = new ConstructorObjectBuilderFactory($factory, $settings->nativeConstructors, $constructors);
-                $factory = new DateTimeObjectBuilderFactory($factory, $constructors);
+                $factory = new DateTimeObjectBuilderFactory(
+                    $factory,
+                    $constructors,
+                    $settings->dateTimeFormats
+                );
                 $factory = new AttributeObjectBuilderFactory($factory);
                 $factory =  new CollisionObjectBuilderFactory($factory);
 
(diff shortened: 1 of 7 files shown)


boesing:
Oh, it seems that I started from a very old `master` branch...

romm:
Hi @boesing sorry for the late answer.

Thank you very much for your contribution, it was indeed needed to be able to customize which date formats were to be supported by the mapper. The funny thing is that just when you submitted this contribution, we had the same issue with a customer, and I started the work to be able to submit #208 (#195 and #207 were needed for that).

Although your proposed solution was interesting, it required another method in the builder, which has already grown quite a lot lately. The solution I'm proposing should be more flexible and more adapted to the library, I'd love to read your feedback if you have time to review it.

>  Until then, thanks for all the work here - this is one of those libraries which had the most impact in my development experience.

Thank you for the kind words, that makes me want to keep going! 🤗

boesing:
Closing in favor of #208

PR #201, opened 2022-08-12
Ensure cached files do not already exist

mtouellette:
seems to be necessary with concurrent requests

### DIFF
--- a/src/Cache/Compiled/CompiledPhpFileCache.php
+++ b/src/Cache/Compiled/CompiledPhpFileCache.php
@@ -86,8 +86,10 @@ final class CompiledPhpFileCache implements CacheInterface
         $tmpFilename = $tmpDir . DIRECTORY_SEPARATOR . uniqid('', true);
         $filename = $this->path($key);
 
-        if (! @file_put_contents($tmpFilename, $code) || ! @rename($tmpFilename, $filename)) {
-            throw new CompiledPhpCacheFileNotWritten($filename);
+        if (!file_exists($filename)) {
+            if (!@file_put_contents($tmpFilename, $code) || !@rename($tmpFilename, $filename)) {
+                throw new CompiledPhpCacheFileNotWritten($filename);
+            }
         }
 
         return true;
--- a/src/Cache/Compiled/CompiledPhpFileCache.php
+++ b/src/Cache/Compiled/CompiledPhpFileCache.php
@@ -86,10 +86,16 @@ final class CompiledPhpFileCache implements CacheInterface
         $tmpFilename = $tmpDir . DIRECTORY_SEPARATOR . uniqid('', true);
         $filename = $this->path($key);
 
-        if (!file_exists($filename)) {
-            if (!@file_put_contents($tmpFilename, $code) || !@rename($tmpFilename, $filename)) {
+        try {
+            if (! @file_put_contents($tmpFilename, $code)) {
+                throw new CompiledPhpCacheFileNotWritten($tmpFilename);
+            }
+
+            if (! file_exists($filename) && ! @rename($tmpFilename, $filename)) {
                 throw new CompiledPhpCacheFileNotWritten($filename);
             }
+        } finally {
+            @unlink($tmpFilename);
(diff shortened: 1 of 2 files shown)


romm:
Hi @mtouellette thanks for the contribution. I'm ok with the change, however I'd like to know what was the effect of the issue? Did some error get reported?

mtouellette:
The CompiledPhpCacheFileNotWritten exception is thrown and ultimately caught in the application as a failed request.  When this happens the file already exists and is valid but can't be overwritten (I believe because another request is reading from the file).

romm:
Thanks for the clarification! I added more checks.

Don't know why the rebase did override the original commit author though 🤔

mtouellette:
Hi @romm, thanks for the speedy merge but I think you'll need to add the file exists check on the file_put_contents line.  I ran a quick check and I get the error with the changes that were committed.

mtouellette:
Huh, actually no.  It still errors out even with that check because the file_put_contents still runs.  The file_put_contents shouldn't happen if the file already exists.

romm:
Well this is really strange, because concurrent requests will use unique temporary file names because of:

https://github.com/CuyZ/Valinor/blob/cbec76d599dda64571ebc97d3ab9be338638f4d0/src/Cache/Compiled/CompiledPhpFileCache.php#L86

What is the actual error message you get? Does it contain `.valinor.tmp`?

mtouellette:
Good point but it is failing this way.  Let me break it up a bit and set some break points to see if I can nail this down.

mtouellette:
Ah, the rename is failing  because I'm running on Windows (not my choice).   I think it needs to be more like below.  There may also be an issue with the mkdir call in this same method but I'll need to review that later (maybe tonight or tomorrow).



        try {

            if (!@file_put_contents($tmpFilename, $code)) {
                throw new CompiledPhpCacheFileNotWritten($tmpFilename);
            }

            //handles rename() Access is denied. (code: 5) in Windows
            $renamed = (@rename($tmpFilename, $filename)) ? true : @copy($tmpFilename, $filename);

            if (!@file_exists($filename) && !$renamed) {
                throw new CompiledPhpCacheFileNotWritten($filename);
            }

        } finally {
            
            if(file_exists($tmpFilename)) {
                @unlink($tmpFilename);
            }
        }

romm:
Thanks for digging in. Did you check that the cache directory has the correct permission?

mtouellette:
I propose something like the code below.  The mkdir function returns false if the directory already exists and in the case of Mezzio the `@` doesn't properly suppress the warning thrown when it does.  I think the is_writable check along with the suppression via the error handler is enough to ensure the directory exists.  Thoughts?



    public function set($key, $value, $ttl = null): bool
    {
        $code = $this->compile($value, $ttl);

        $tmpDir = $this->cacheDir . DIRECTORY_SEPARATOR . '.valinor.tmp';

        if(!is_writable($this->cacheDir)) {
            throw new CacheDirectoryNotWritable($this->cacheDir);
        }

        //suppress 
        set_error_handler(function($errno, $errstr = '', $errfile = '', $errline = 0) {
            return true;
        }, E_WARNING);

        mkdir($tmpDir, self::TEMPORARY_DIR_PERMISSION, true);

        restore_error_handler();

        /** @infection-ignore-all */
        $tmpFilename = $tmpDir . DIRECTORY_SEPARATOR . uniqid('', true);
        $filename = $this->path($key);

        try {

            $fileExists = file_exists($filename);

            if (!@file_put_contents($tmpFilename, $code)) {
                throw new CompiledPhpCacheFileNotWritten($tmpFilename);
            }

            //handles rename() Access is denied. (code: 5) in Windows
            $renamed = (rename($tmpFilename, $filename)) ? true : copy($tmpFilename, $filename);

            if (!$fileExists && !$renamed) {
                throw new CompiledPhpCacheFileNotWritten($filename);
            }

        } finally {

            if(file_exists($tmpFilename)) {
                @unlink($tmpFilename);
            }
        }


        return true;
    }

PR #208, opened 2022-08-30
feat!: introduce constructor for custom date formats

romm:
A new constructor can be registered to declare which format(s) are
supported during the mapping of a date object. By default, any valid
timestamp or ATOM-formatted value will be accepted.

```php
    // Both COOKIE and ATOM formats will be accepted
    ->registerConstructor(
        new \CuyZ\Valinor\Mapper\Object\DateTimeFormatConstructor(DATE_COOKIE, DATE_ATOM)
    )
    ->mapper()
    ->map(DateTimeInterface::class, 'Monday, 08-Nov-1971 13:37:42 UTC');
```

The previously very opinionated behaviour has been removed, but can be
temporarily used to help with the migration.

```php
    ->registerConstructor(
        new \CuyZ\Valinor\Mapper\Object\BackwardCompatibilityDateTimeConstructor()
    )
    ->mapper()
    ->map(DateTimeInterface::class, 'Monday, 08-Nov-1971 13:37:42 UTC');
```


boesing:
Hey @romm,
so you are right. Having this as a feature would work. Thats true. What I usually don't like is when I have to instantiate stuff in userland. Configuring is welcome as I dont have to hassle with implementation Details and refactoring of the component can be easily done without enforcing userland code to change.

With this constructor stuff, I do have to know deep internals of this component which makes it kinda complicated. 

Is there a reason why you try to avoid configuration and instead require userland to instantiate objects? 
I would be fine with a configuration object where I can pass my `valinor.config.php` or whatever and then passing it to the builder. 

But this constructor detail is kinda specific to how valinor works in detail and as a consumer I do not want to dive too deep into the functionality of a component.

boesing:
If the builder starts growing, adding a configuration object might help which then is passed to the builder constructor. This way, the builder keeps clean and only consumes configuration. 🤷🏼‍♂️

romm:
Right, thank you for your valuable feedback!

What do you think of the last commit? Does the method's name look ok to you?

src/Mapper/Object/Exception/CannotParseToDateTime.php
```diff
@@ -13,18 +13,20 @@
...
 {
-    private string $body = 'Value {value} does not match a valid date format.';
+    private string $body = 'Value {value} does not match any of the following formats: {formats}.';
 
     /** @var array<string, string> */
     private array $parameters;
 
     /**
      * @param string|int $datetime
+     * @param non-empty-list<string> $formats
```
boesing, line 23: shouldn't this list contain only `non-empty-string`?

boesing (approved):
LGTM, just one thing but not a blocker tho.

So this is "best" of both worlds I guess. Having a clean configuration while still providing the ability to add custom stuff if needed. So thats fine for me. I'll close my PR in favor of this one.

Thanks!

src/Mapper/Object/Exception/CannotParseToDateTime.php
romm: Good catch! I will also update it in the actual constructors.

romm:
Thanks for the review, and for the initial PR submission,  @boesing!

src/Mapper/Object/DateTimeFormatConstructor.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * ```php
+ * (new \CuyZ\Valinor\MapperBuilder())
+ *     // Both `Cookie` and `ATOM` formats will be accepted
+ *     ->registerConstructor(new DateTimeFormatConstructor(DATE_COOKIE, DATE_ATOM))
+ *     ->mapper()
+ *     ->map(DateTimeInterface::class, 'Monday, 08-Nov-1971 13:37:42 UTC');
+ * ```
+ *
+ * @internal
```
Slamdunk, line 29: Hi, may I ask you why is it `@internal`?

`MapperBuilder::supportDateFormats` is great, but I'd rather rely on `registerConstructor(new [...])` api which seems to me more consistent and explicit

romm: Hi @Slamdunk, this is mostly because I agreed with https://github.com/CuyZ/Valinor/pull/208#issuecomment-1231735731, honestly I didn't think people would prefer using the actual constructor object over the `MapperBuilder::supportDateFormats()` method.

> […] which seems to me more consistent and explicit

I agree it can be more consistent, however is it really more explicit? 🤔 

If this is really wanted, I could add it to the API scope. I'll give it some thoughts.

boesing: I use this as a library which I want to configure. I do not want to care about internals. In the end, 99% of my user experience is to call the `map` method. 

If this library will be rewritten in the future - as it was already the case for at least 1-3 times, I will always end up refactoring my code because I used internal details like a dateformatter or whatever. But I do not care HOW it is resolved in detail, I just want to configure my settings. 

I like the way it was resolved right now. Since the constructor stuff seems to be a thing, I would be also fine with removing the internal annotation. 

So both worlds can use it how they want. I did not even notice the internal in the PR as, as I mentioned, dont care about how it is resolved 😄

boesing: Oh and btw - I do understand my dependency but I hate it when I have to instantiate the whole thing. 

Imagine there wont be a configuration for doctrine and you would be required to instantiate all the dependencies. Hell I would drop that thing.

Slamdunk: > Imagine there wont be a configuration for doctrine and you would be required to instantiate all the dependencies. Hell I would drop that thing.

I indeed instantiate Doctrine and don't use its configuration because the latter always hit hard on my projects, required me do dive in the docs **after** the new major for hours and made me lose time without adding value.

`new A(new B, new C)` always gave me a faster upgrade path: it requires reading less docs because as soon as a class gets deprecated I get noticed by Static Analysis and immediately change the objects before the next major.

PR #216, opened 2022-09-15
Fix psr/simple-cache deprecations

lookyman:
Fixes `User Deprecated: Method "Psr\SimpleCache\CacheInterface::get()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "CuyZ\Valinor\Cache\..." now to avoid errors or add an explicit @return annotation to suppress this message.` reported by `symfony/error-handler`.


romm:
Thanks @lookyman!

PR #219, opened 2022-09-21
fix: handle constructors that are static anonymous functions

edudobay:
The `MethodObjectBuilder` was incorrectly used when a registered constructor is a static anonymous functions — it was handled like a static method closure `Class::method(...)` and would yield errors like this:

    Error: Call to undefined method stdClass::CuyZ\Valinor\Tests\Integration\Mapping\{closure}()

PHP Reflection does not provide any way of telling static functions and closures of static methods apart, other than checking for the name `{closure}`. We check that `{closure}` is actually the last part of the fully-qualified name, instead of just checking that the string ends with `{closure}`.

### DIFF
--- a/src/Definition/FunctionDefinition.php
+++ b/src/Definition/FunctionDefinition.php
@@ -22,6 +22,8 @@ final class FunctionDefinition
 
     private bool $isStatic;
 
+    private bool $isClosure;
+
     private Parameters $parameters;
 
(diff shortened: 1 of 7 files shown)


romm:
Thanks @edudobay, good catch and nice PR!

FYI I moved the closure detection inside the function definition (which is kind of an upgraded reflection class).

PR #220, opened 2022-09-24
feat: import the `TokenParser` from `doctrine/annotations`

Mopolo:


romm:
Thanks for the work here!

PR #226, opened 2022-09-30
Switch to using getAtomicTypes

danog:
`getAtomicTypes` should be used to iterate over the types of a Union, removing the call to `getChildNodes` since it's out of place and may be deprecated in psalm v5.


danog:
See https://github.com/vimeo/psalm/pull/8525, it should get merged once I finish the other tasks listed in the description, `getChildNodes` is replaced with `getChildNodeKeys` to allow direct substitution of properties.

The main issue with how you're using `getChildNodes` is that it should only really be used by Psalm's NodeIterator (now TypeIterator) classes, in all other cases where the exact type is known we have access to either getter methods (`getAtomicTypes`) or properties (Atomics) that can be directly used.

romm:
Perfect, thank you!

PR #242, opened 2022-10-25
misc: do not use `uniqid()`

TimWolla:
The return value of `uniqid()` is not very random and easily guessable. While this likely does not introduce any issues in CompiledPhpFileCache, the fact that `uniqid()` is often mis-used in locations where actual randomness is a hard requirement, makes this a red flag.

Replace the use of `uniqid()` to generate a random filename by hexadecimal encoded `random_bytes(16)`, giving 128 Bits of randomness, making the value unguessable in practice and preventing the `CompiledPhpFileCache` from showing up when searching for `uniqid()`.

### DIFF
--- a/src/Cache/Compiled/CompiledPhpFileCache.php
+++ b/src/Cache/Compiled/CompiledPhpFileCache.php
@@ -16,14 +16,15 @@ use Psr\SimpleCache\CacheInterface;
 use Traversable;
 
 use function assert;
+use function bin2hex;
 use function file_exists;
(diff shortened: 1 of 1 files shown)


romm:
Thank you both for the contribution and the explanation. 🙂

TimWolla:
> and the explanation. slightly_smiling_face

You’re welcome. I strive to make my commit messages as useful as possible. Thanks for creating Valinor!

PR #243, opened 2022-10-25
Add additional `export-ignore` files to `.gitattributes`

TimWolla:
The lock file is not useful for downstream consumers and requires more than 200kB of data to be unnecessarily transmitted.

---------------

Also: Is the `/docs/` directory useful within the `vendor/` directory or can that one also be ignored?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
 *.php text eol=lf
 
 /.github/ export-ignore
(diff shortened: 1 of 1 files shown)


romm:
I guess we can ignore the `docs` folder as well, would you mind adding it to the list?

romm:
That was fast 🚀 

Thank you!

PR #246, opened 2022-11-05
feat: add support for `strict-array` type

danog:
This PR adds basic support for strict array tokens, just so I can keep the new section in the Psalm docs mentioning Valinor as a way to validate strict arrays :)

Phpstan should also support strict array syntax once https://github.com/phpstan/phpdoc-parser/pull/161 is merged.


romm:
Hi @danog, thanks for the contribution, some tests are missing but we can add them later — I'm preparing a release now to help with vimeo/psalm#8683

danog:
Thank you, very much appreciated :D

PR #253, opened 2022-11-15
misc: remove `@` from comments for future PHP versions changes

agustingomes:
Context:

As currently stands for the 1.14.x branch of the `doctrine/annotations`, code units with `@PHP8` annotations fail to be loaded.

This happens because the default list of implicitly ignored annotations does not contain `PHP8` in it, and therefore in scenarios where applications are loaded with components with such annotations fail with errors of this nature:

```sh
PHP Fatal error:  Uncaught Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@PHP8" in method CuyZ\Valinor\Cache\FileSystemCache::get() was never imported. Did you maybe forget to add a "use" statement for this annotation? in vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php:40
```

This change introduces the `TODO` annotation which is implicitly ignored by default, which helps avoid applications running into issues because of the annotation not being loaded.

### DIFF
--- a/src/Cache/ChainCache.php
+++ b/src/Cache/ChainCache.php
@@ -30,7 +30,7 @@ final class ChainCache implements CacheInterface
     }
 
     /**
-     * @PHP8.0 add `mixed` return type and remove PHPDoc
+     * PHP8.0 add `mixed` return type and remove PHPDoc
      *
      * @return EntryType|null
      */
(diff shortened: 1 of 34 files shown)


romm:
Thank you!

Could you just, instead of replacing `@PHPX` with `@TODO PHPX`, remove the `@` so it becomes `PHPX`? I use `@todo` annotations a lot during development, I search and fix them before committing my work, so I prefer not to have already existing entries polluting the codebase. 😛

agustingomes:
will do 👍🏼

romm:
Thank you!

PR #260, opened 2022-11-22
misc!: mark tree mapper and arguments mapper as `@pure`

romm:
This change also impacts all callables that can be given to the mapper builder — they also need to be pure.

Although it reduces the scope of the mappers, it greatly improves their safety.


Ocramius (approved):
:muscle:

PR #271, opened 2022-11-29
Allow for timestamps before the epoch

markhuot:
I'm not sure this is the _right_ fix but timestamps can technically be negative numbers. I have some dates that I'm importing that range back to 1920 and am currently getting an error that `Value {source_value} does not match any of 'non-empty-string', 'positive-int'` for the value `-450734400`.

If you agree with this change I can expand the PR, update the tests, etc.

### DIFF
--- a/src/Mapper/Object/DateTimeFormatConstructor.php
+++ b/src/Mapper/Object/DateTimeFormatConstructor.php
@@ -44,7 +44,7 @@ final class DateTimeFormatConstructor
 
     /**
      * @param class-string<DateTime|DateTimeImmutable> $className
-     * @param non-empty-string $value
+     * @param non-empty-string|int $value
      */
     #[DynamicConstructor]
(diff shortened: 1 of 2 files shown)


romm:
Hi @markhuot, thank you for the suggestion; I didn't know about this usage of the timestamp.

It makes sense to me: please add tests to the PR, and I will be happy to merge it. 🙂

src/Mapper/Object/DateTimeFormatConstructor.php
```diff
@@ -44,7 +44,7 @@ public function __construct(string $format, string ...$formats)
 
     /**
      * @param class-string<DateTime|DateTimeImmutable> $className
-     * @param non-empty-string|positive-int $value
+     * @param non-empty-string $value
```
romm, line 47:
```suggestion
     * @param non-empty-string|int $value
```

romm:
Thank you guys!

PR #272, opened 2022-12-01
misc: improve return type signature for `TreeMapper`

vvoody-:
In situation when $signature is just type of class-string without exact class, return type expected by phpstan is mixin. With this change phpstan will expect return type object.

Bit more context here https://github.com/phpstan/phpstan/issues/8443

### DIFF
--- a/src/Mapper/TreeMapper.php
+++ b/src/Mapper/TreeMapper.php
@@ -16,7 +16,7 @@ interface TreeMapper
      * @return (
      *     $signature is class-string<T>
      *         ? T
-     *         : mixed
+     *         : ($signature is class-string ? object : mixed)
      * )
      *
      * @throws MappingError


romm:
Hi @vvoody- thanks for the contribution 🙂 

Could you post an example to explain why this change would make sense?

romm:
Oops I just saw you posted https://phpstan.org/r/7d4b1d4c-0f9d-4906-867c-c893dfdf7de3

I can see places where it can help indeed, although it is always preferred to use a class-string of a given object sometimes we just can't.

Thanks!

vvoody-:
When I think about it now, its hard to justify this change by my needs :) because its caused by magic in my project (or rather in one library I'm using). Basically I have class (controller) with methods (api endpoints) having attribute with information to what class should be request body mapped to with Valinor. There is one ugly drawback to it - I have to inline annotate constructed object so now there is its class name twice in this code.
```php
	#[Path('/by-zones')]
	#[Method('POST')]
	#[RequestBody(entity: OfferRequestByZones::class)]
	public function byZone(ApiRequest $apiRequest, ApiResponse $apiResponse): ApiResponse
	{
		/** @var OfferRequestByZones $entity */
		$entity = $apiRequest->getEntity();

		// ...
	}
	
```

Code where I use Valinor for actual mapping is separated and there is no information about actual class in `$signature` for static analysis. It have only type `class-string` there. That made phpstan expect return value from map function to be mixin but it is actually always object in my case. Temporarily fixed by another ugly inline var annotation
```php
/** @var object $entity */
$entity = $this->treeMapper->map($entityClass, $request->getJsonBody());
```

romm:
I see, thanks for the detailed explanation.

BTW did you know Valinor can [map arguments of a callable](https://valinor.cuyz.io/latest/how-to/map-arguments-of-a-callable/)? Don't know if that can apply there, just in case it can help… 🙂

vvoody-:
Hmm I can see it being useful. I can define my DAO representing request body as third argument of my byZone method. That would need some rewriting of rest api library I'm using, but also it must be possible to tell Valinor to do mapping only for third parameter. At that point I can instead just grab that one typehint of third parameter by myself with reflection and use just TreeMapper. And that is actually good idea how can I get rid of that inline annotation.

romm:
> but also it must be possible to tell Valinor to do mapping only for third parameter

Actually no, the current behavior is to map all arguments of the given callable. Mapping specific arguments could be an evolution of the feature, though. I'll have to think about it.

PR #322, opened 2023-01-09
Improve TODO handling

szepeviktor:
- safe YAML literal style
- using 1 command
- find FIXME-s
- whole words only

cons: no output on success just a ✔️

### DIFF
--- a/.github/workflows/quality-assurance.yml
+++ b/.github/workflows/quality-assurance.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Checking for forgotten TODO
         run: |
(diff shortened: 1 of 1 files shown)


szepeviktor:
Please test it a bit. I may be wrong!

romm:
Didn't know about the `git grep` command, thanks!

I disabled the word boundaries option in the regex, because I sometime can use names like `TodoService` during development, which would not be catch with the `\b` option; but that could have been useful, thanks for that!

szepeviktor:
You're welcome.
I'm very happy that my brainchildren are useful for others.

PR #357, opened 2023-03-08
Set the `$previous` Exception in various custom Exceptions

TimWolla:
This was already done for some exceptions, but not all of them. The second commit is separate, because I found it less obviously to be correct. I can squash it into the first one if you want.


romm:
Hi @TimWolla, thank you for the contribution.

Before merging, I'd like to understand in which cases this change is useful, have you had a use-case where you needed the previous exception? These exceptions you changed are marked `@internal`, but I guess you can still catch native `\Exception` and check for previous, is it how you use it?

TimWolla:
> Before merging, I'd like to understand in which cases this change is useful, have you had a use-case where you needed the previous exception?

I don't remember the exact example, but it was a case where I've specified some invalid type signature and the error message in `InvalidMappingTypeSignature` was not helpful; and neither was the stack trace, because it just pointed to this line:

https://github.com/CuyZ/Valinor/blob/eebd3141d89b1a5a7b6bd5388159bf736c89c8c6/src/Mapper/TypeTreeMapper.php#L40

Having the `$previous` exception available would have been helpful to me as a human debugger, because it would've provided me the necessary context (error message + stack trace) to find the issue in my type signature. In fact I've patched my local `vendor/` folder to store `$previous` for debugging.

In other words: I'm not interested in `$previous` for programmatic consumption. I'm interested in it, because it provides useful context for the human reader if an uncaught Exception (invalid type signature is not something that should be caught) makes it into the Exception pretty printer. If a future refactoring changes the types of the internal exceptions, then that's fine, because the code itself doesn't care.

PR #369, opened 2023-04-06
fix: prevent infinite loop when class has parent class with same name

### LINKED ISSUE #352: Segfault during custom date time object
Hi,
     it looks like infinite loop when `AdvancedClassNameToken `try to resolve parent type `\DateInterval`. When I use custom date time object with custom date interval object everything after `CustomDateTime::diff` method goes into infinite loop. I am using valinor 1.3.0 but it was starting from version 1.1.0.

**Reproduction**:
```php

final class Request {
    public function __construct(
        public readonly CustomDateTime $time,
    ) {}
}

final class CustomDateTime extends \DateTimeImmutable
{
    /**
     * @throws Exception
     */
    public function diff($targetObject, bool $absolute = false): CustomeDateInterval 
    {
        return new CustomeDateInterval ('P1Y');
    }
}

final class CustomeDateInterval extends \DateInterval {

}


//segmentation fault
$mapper = (new MapperBuilder())->mapper();
$mapper->map(Request::class, ['time' => '']);
```

**Expected behavior**:
Valinor can map `CustomeDateInterval ` with `CustomDateTime`

**UPDATE:**

It is really because of `CustomDateTime::diff` return `CustomDateInterval`.
This implementation of `diff` method with default `DateInterval` works fine
```php
public function diff($targetObject, bool $absolute = false): \DateInterval
    {
        return new \DateInterval ('P1Y');
    }
```

pauci:
Attempts to fix #352.

### DIFF
--- a/src/Type/Parser/Lexer/AliasLexer.php
+++ b/src/Type/Parser/Lexer/AliasLexer.php
@@ -91,6 +91,10 @@ final class AliasLexer implements TypeLexer
             return $symbol;
         }
 
+        if (Reflection::classOrInterfaceExists($symbol)) {
+            return $symbol;
+        }
+
         $full = $namespace . '\\' . $symbol;
 
         if (Reflection::classOrInterfaceExists($full)) {
--- a/src/Type/Parser/Lexer/AliasLexer.php
+++ b/src/Type/Parser/Lexer/AliasLexer.php
@@ -32,6 +32,10 @@ final class AliasLexer implements TypeLexer
 
     private function resolve(string $symbol): string
     {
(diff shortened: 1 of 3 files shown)


romm:
Excellent, thank you!

hydrapolic:
@romm would it be possible to create a new release so we can pin to a version instead of dev-master in our composer? :)

Thanks!

romm:
You got it!

https://github.com/CuyZ/Valinor/releases/tag/1.4.0

Enjoy. 😉

hydrapolic:
Thank you @romm!

PR #373, opened 2023-04-17
misc: catch `json_encode` exception to help identifying parsing errors

TimWolla:
This brings two improvements:

- The reason for the JSON parsing failure is visible in the stack trace / $previous exception, allowing the developer to determine whether the parsing failure was caused by a syntax error, invalid UTF-8 sequence or something else.
- The JSON string `null` (which decodes to `null`) is no longer erroneously rejected as invalid. Instead it is rejected, because it is not iterable, slightly improving the error reporting.

### DIFF
--- a/src/Mapper/Source/JsonSource.php
+++ b/src/Mapper/Source/JsonSource.php
@@ -9,10 +9,13 @@ use CuyZ\Valinor\Mapper\Source\Exception\InvalidSource;
 use CuyZ\Valinor\Mapper\Source\Exception\SourceNotIterable;
 use Iterator;
 use IteratorAggregate;
+use JsonException;
 use Traversable;
 
 use function is_iterable;
 use function json_decode;
+
+use const JSON_THROW_ON_ERROR;
 
 /**
  * @api
@@ -29,10 +32,10 @@ final class JsonSource implements IteratorAggregate
      */
     public function __construct(string $jsonSource)
     {
-        $source = json_decode($jsonSource, true);
-
-        if ($source === null) {
-            throw new InvalidJson($jsonSource);
+        try {
+            $source = json_decode($jsonSource, true, flags: JSON_THROW_ON_ERROR);
+        } catch (JsonException $e) {
+            throw new InvalidJson($jsonSource, $e);
         }
 
         if (! is_iterable($source)) {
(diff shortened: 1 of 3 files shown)


src/Mapper/Source/JsonSource.php
```diff
@@ -29,10 +32,10 @@ final class JsonSource implements IteratorAggregate
      */
     public function __construct(string $jsonSource)
     {
-        $source = json_decode($jsonSource, true);
-
-        if ($source === null) {
-            throw new InvalidJson($jsonSource);
+        try {
+            $source = json_decode($jsonSource, true, 512, JSON_THROW_ON_ERROR);
```
TimWolla, line 36: The escaped mutant is also a good reason for named parameters here. I am happy to make the necessary change if named parameters are acceptable to you.

TimWolla:
@romm Ping. Changed to a named argument now.

src/Mapper/Source/JsonSource.php
```diff
@@ -29,10 +32,10 @@ final class JsonSource implements IteratorAggregate
      */
     public function __construct(string $jsonSource)
     {
-        $source = json_decode($jsonSource, true);
-
-        if ($source === null) {
-            throw new InvalidJson($jsonSource);
+        try {
+            $source = json_decode($jsonSource, true, flags: JSON_THROW_ON_ERROR);
```
romm, line 36:
```suggestion
            $source = json_decode($jsonSource, associative: true, flags: JSON_THROW_ON_ERROR);
```

While we're there, we should ease the reading for `associative` as well. 🙂

TimWolla:
Made another small fix I just noticed: `use JsonException` instead of using `\JsonException`. Wasn't caught by CI apparently.

> That was fast!

For my open source contributions I generally try to act on any review requests quickly, so that the maintainer doesn't need to wait for me.

src/Mapper/Source/JsonSource.php
TimWolla: Okay, done. Let me know if you prefer splitting the function call across multiple lines (and if "yes" whether or not to use a trailing comma). Also let me know if I should squash my commits.

romm: That's very good like it is, thanks for asking. I'll do the squashing myself!

TimWolla:
Merci!

PR #376, opened 2023-04-25
Fix mapping of unions of scalars and classes (fixes #375).

### LINKED ISSUE #375: Regression: Union of scalar and class
It used to be possible to map a type like `string|Foobar`, but with recent releases this now causes an error when trying to map `Foobar`:

```php
class Foobar
{
    public function __construct(public int $foo, public string $bar) {}
}

var_dump(
    (new CuyZ\Valinor\MapperBuilder())
        ->mapper()
        ->map("list<string|Foobar>", ["foo", ["foo" => 123, "bar" => "baz"], "bar"])
);
```

Expected output:

```
array(3) {
  [0]=>
  string(3) "foo"
  [1]=>
  object(Foobar)#96 (2) {
    ["foo"]=>
    int(123)
    ["bar"]=>
    string(3) "baz"
  }
  [2]=>
  string(3) "bar"
}
```

Actual output:

```
PHP Fatal error:  Uncaught CuyZ\Valinor\Mapper\TypeTreeMapperError: Could not map type `list<string|Foobar>`. An error occurred at path 1: Invalid value array{foo: 123, bar: 'baz'}. in Valinor/src/Mapper/TypeTreeMapper.php:29
```

`git bisect` points to 72cba320f582c7cda63865880a1cbf7ea292d2b1 as the cause.

AndrolGenhald:
Modify `UnionNodeBuilder` to try building objects even if not all of the types in the union are classes. Scalars are still preferred if both a scalar and class could match the value since `CasterProxyNodeBuilder` will check if the `UnionType` accepts the scalar before `UnionNodeBuilder` is tried.


romm:
Hi @AndrolGenhald, sorry it took that long. Thanks a lot for the contribution and the well-tested behavior! 💪

PR #378, opened 2023-05-04
Avoid using unserialize when caching NULL default values

danog:
`is_scalar` should just be an inverted `is_object` check, instead.


romm:
Hi @danog, thank you! As this PR comes from an organization I don't have write access on it, could you rebase your branch on master please? 🙂

romm:
Thanks @danog!

Issue #379, opened 2023-05-16
Psalm plugin failed to load


rubenrubiob:
Hello,

When trying to use the Psalm plugin as [stated in the documentation](https://valinor.cuyz.io/latest/other/static-analysis/), I get the following error:

```
Uncaught UnexpectedValueException: CuyZ\Valinor\QA\Psalm\ValinorPsalmPlugin is not a known class in /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Config.php:1637
Stack trace:
#0 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Config.php(1518): Psalm\Config->loadPlugin()
#1 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(244): Psalm\Config->processPluginFileExtensions()
#2 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(335): Psalm\Internal\Analyzer\ProjectAnalyzer->__construct()
#3 /var/www/valinor-sample/vendor/vimeo/psalm/psalm(9): Psalm\Internal\Cli\Psalm::run()
#4 /var/www/valinor-sample/vendor/bin/psalm(120): include('...')
#5 {main}

Next Psalm\Exception\ConfigException: Failed to load plugin CuyZ\Valinor\QA\Psalm\ValinorPsalmPlugin in /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Config.php:1647
Stack trace:
#0 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Config.php(1518): Psalm\Config->loadPlugin()
#1 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(244): Psalm\Config->processPluginFileExtensions()
#2 /var/www/valinor-sample/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(335): Psalm\Internal\Analyzer\ProjectAnalyzer->__construct()
#3 /var/www/valinor-sample/vendor/vimeo/psalm/psalm(9): Psalm\Internal\Cli\Psalm::run()
#4 /var/www/valinor-sample/vendor/bin/psalm(120): include('...')
#5 {main}
[...]

PR #380, opened 2023-05-28
Move QA from autoload-dev to autoload

BenMorel:
Attempting to use the Psalm plugin yields the following error:

> CuyZ\Valinor\QA\Psalm\ValinorPsalmPlugin is not a known class

This is because classes registered as `autoload-dev` are only included in the autoloader when running composer in dev mode **on the Valinor project**, not when Valinor is a project dependency.

Moving it to `autoload` will make it available everywhere.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,13 +33,13 @@
     },
     "autoload": {
         "psr-4": {
-            "CuyZ\\Valinor\\": "src"
+            "CuyZ\\Valinor\\": "src",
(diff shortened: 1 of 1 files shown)


romm:
Hi @BenMorel, thanks for the contribution however this doesn't seem to be a good solution to include quality-assurance files in production.

I wrote a temporary solution for this issue there: https://github.com/CuyZ/Valinor/issues/379#issuecomment-1570019999; let's hope we find a better way in the future.

PR #402, opened 2023-08-17
fix: crash when trying to resolve methods from abstract class with interface and with method referencing self

simPod:
I have encountered the issue when working with brick/math, namely `BigNumber::of()`.


romm:
Thanks @simPod!

PR #408, opened 2023-08-19
fix: load attributes lazily during runtime and cache access

### LINKED ISSUE #393: Attributes in cache files?
Hey there,

maybe I am missing something, but the cache files do also contain informations for attributes. 
I am not 100% sure why valinor needs attributes as there is no attribute support regarding type addition from what I can see?

We recently ran into an issue where our OpenAPI Attributes of a request object were persisted in the cache since we do generate the cache at build time while dev dependencies are still available. We changed this to non-dev composer dependencies now but I still wonder why attributes are even parsed at all.

romm:
Attributes are now loaded in a lazy way, preventing an error from happening when attributes were compiled in the cache, but not present when using the mapper during the runtime.

This could happen for instance with attributes used to generate documentation of an API: they are loaded as a development dependency, then the cache is warmed up during deployment, so they are present in the mapper compiled cache; but these attributes are missing from the live version of the application, so they cannot be accessed.

This commit helps to prevent a fatal error from happening in this kind of situation.

Fixes #393

### DIFF
--- a/src/Definition/AttributesContainer.php
+++ b/src/Definition/AttributesContainer.php
@@ -7,18 +7,28 @@ namespace CuyZ\Valinor\Definition;
 use Traversable;
 
 use function array_filter;
+use function array_map;
 use function array_values;
 use function count;
+use function is_a;
 
-/** @internal */
+/**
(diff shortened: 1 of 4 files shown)


romm:
Thanks for the review!

PR #409, opened 2023-08-22
feat: Support microseconds in the Atom / RFC 3339 / ISO 8601 format

TimWolla:
This commit adds default support for microseconds in the RFC 3339 format.

The references to the “Atom” format have been updated to RFC 3339 which is identical to the Atom format, but uses a reference that is not PHP-specific.

### DIFF
--- a/tests/Integration/Mapping/Object/DateTimeMappingTest.php
+++ b/tests/Integration/Mapping/Object/DateTimeMappingTest.php
@@ -24,7 +24,7 @@ final class DateTimeMappingTest extends IntegrationTest
         }
     }
 
-    public function test_default_date_constructor_with_valid_atom_format_source_returns_datetime(): void
+    public function test_default_date_constructor_with_valid_rfc_3339_format_source_returns_datetime(): void
     {
         try {
             $result = (new MapperBuilder())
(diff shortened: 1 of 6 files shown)


TimWolla:
Failures appear to be unrelated and probably depend on the ICU version. It's a little odd that the error code appears to be `U_ZERO_ERROR` which indicates that no error happened. Unfortunately I'm unable to reproduce this locally.

TimWolla:
Test failures are a PHP regression: php/php-src#12020

romm:
Hi @TimWolla, thanks!

I'd like more input on the reasons why you think this particular format should be available. I have to be honest, I did chose to accept `DATE_ATOM` format by default for no real reason, mostly because from what I saw it was often used in other popular projects (can't even remember which ones).

Is there some kind of general approving on the RFC 3339 format in the PHP/web/software development community?

tests/Integration/Mapping/Object/DateTimeMappingTest.php
```diff
@@ -100,7 +130,7 @@ public function test_default_date_constructor_with_invalid_source_throws_excepti
             $error = $exception->node()->messages()[0];
 
             self::assertSame('1630686564', $error->code());
-            self::assertSame("Value 'invalid datetime' does not match any of the following formats: `Y-m-d\TH:i:sP`, `U`.", (string)$error);
+            self::assertSame("Value 'invalid datetime' does not match any of the following formats: `Y-m-d\TH:i:sP`, `Y-m-d\TH:i:s.uP`, `U`.", (string)$error);
```
TimWolla, line 133: You can see in this change that the format didn't actually change. There is just the new `Y-m-d\TH:i:s.uP` which differs by the addition of `.u` for microseconds.

This new format handles the output of JavaScript's `Date.prototype.toISOString()`:

```
> (new Date()).toISOString()
'2023-08-22T18:14:31.112Z'
```

romm:
Awesome, thanks for the detailed explanation!

romm:
@TimWolla I can't push to the repository because it is from the WoltLab org; could you rebase on master?

TimWolla:
Yeah, this is a dumb limitation from GitHub. Now done.

romm:
Thanks!

PR #410, opened 2023-08-22
qa: Fix SourceTest without YAML extension

TimWolla:
Previously when the YAML extension was not installed, the test was skipped /
errored entirely, as the data provider is invalid:

> 1) Error
> The data provider specified for CuyZ\Valinor\Tests\Integration\Mapping\Source\SourceTest::test_sources_are_mapped_properly is invalid.
> CuyZ\Valinor\Mapper\Source\Exception\YamlExtensionNotEnabled: The PHP YAML extension is not enabled.
> /pwd/src/Mapper/Source/YamlSource.php:38
> /pwd/src/Mapper/Source/Source.php:55
> /pwd/tests/Integration/Mapping/Source/SourceTest.php:74

Now the YAML source will simply not be provided.

### DIFF
--- a/tests/Integration/Mapping/Source/SourceTest.php
+++ b/tests/Integration/Mapping/Source/SourceTest.php
@@ -70,9 +70,11 @@ final class SourceTest extends IntegrationTest
             Source::json('{"someValue": {"someNestedValue": "foo", "someOtherNestedValue": "bar"}, "someOtherValue": {"someNestedValue": "foo2", "someOtherNestedValue": "bar2"}}'),
         ];
 
-        yield 'YAML' => [
-            Source::yaml("someValue:\n someNestedValue: foo\n someOtherNestedValue: bar\nsomeOtherValue:\n someNestedValue: foo2\n someOtherNestedValue: bar2"),
-        ];
+        if (extension_loaded('yaml')) {
+            yield 'YAML' => [
+                Source::yaml("someValue:\n someNestedValue: foo\n someOtherNestedValue: bar\nsomeOtherValue:\n someNestedValue: foo2\n someOtherNestedValue: bar2"),
+            ];
+        }
 
         yield 'Camel case' => [
             Source::array([


TimWolla:
Hmm. I expected the `@throws` tag to fix the PHPStan issue based on:

https://phpstan.org/blog/bring-your-exceptions-under-control#inline-%40throws

Will need to look into this later, unless you have a good suggestion.

romm:
I think it's ok to not mark the test as skipped when the `yaml` extension is not loaded (skipped tests don't block the test-suite anyway), so just add a `extension_loaded('yaml')` condition on the yield and we're good to go.

TimWolla:
> so just add a `extension_loaded('yaml')` condition on the yield and we're good to go.

Okay, that works for me. Now done (and also updated the commit message + PR description).

tests/Integration/Mapping/Source/SourceTest.php
```diff
@@ -5,22 +5,28 @@
...
 final class SourceTest extends IntegrationTest
 {
     /**
      * @dataProvider sources_are_mapped_properly_data_provider
      *
-     * @param iterable<mixed> $source
+     * @param iterable<mixed>|Throwable $source
      */
-    public function test_sources_are_mapped_properly(iterable $source): void
+    public function test_sources_are_mapped_properly(iterable|Throwable $source): void
```
romm, line 24: All of that is not needed anymore. 😉

TimWolla: :facepalm: Sorry. I only reran PHPUnit locally to make sure it still works, but didn't double check the diff.

romm: No worries!

romm:
Thanks @TimWolla!

tests/Integration/Mapping/Source/SourceTest.php
romm: Btw you can run `composer check` locally, it runs all the QA tools!

romm: Totally understand; thanks again for your dedication!

PR #411, opened 2023-08-22
fix: Fix the handling of invalid formats if `intl.use_exceptions=1`

TimWolla:
If `intl.use_exceptions=1`, ext/intl will throw an `\IntlException` that was not yet handled and caused `StringFormatter::format()` to violate its contract to throw `StringFormatterError` in case of errors.


romm:
Didn't know about that, thanks!

TimWolla:
Yeah, I also learned about this today, while investigating the PHP regression (php/php-src#12020) …

PR #412, opened 2023-08-22
qa: Use less-specific exception message assertion in StringFormatterTest

TimWolla:
The actual error message is emitted by ICU / PHP which we do not have control over.  We should not test that the upstream dependency returns stable error messages, especially since error messages are for human consumption and thus not part of the backwards compatibility contract.

see php/php-src#12020


romm:
Thanks @TimWolla!

TimWolla:
Thanks, rebased the other PRs.

PR #414, opened 2023-08-24
test: Add `@requires` to NodeMessageTest::test_custom_locale_is_used()

TimWolla:
This test requires the intl extension instead of the shim to work.


romm:
Thanks!

PR #416, opened 2023-08-24
chore: Upgrade rector

TimWolla:
This is required for PHP 8.3 support.


TimWolla:
Green, except for the PHP 8.3 tests that have been enabled too early :smile:

romm:
> Green, except for the PHP 8.3 tests that have been enabled too early 😄

Woops!

Thanks 😁

PR #422, opened 2023-08-28
Fix docblock type parsing from variadic params

simPod:
~Hello, I have reproduced in tests a failure when type parser incorrectly parses variadic parameter. There should not be any error but TypesDoNotMatch exception is thrown.~


romm:
Hi @simPod, I've moved the logic to the type resolver where it should belong. Thanks for the report!

simPod:
Nice. One more reason to always include tests 🙃

Issue #394, opened 2023-08-02
LogicException: Namespace not found


JanMikes:
Hi, seems that token parser does not expect namespace without backslash - when trying to map nested  classes in same (single level) namespace, i get `Namespace not found` exception.

```php
namespace Acme;

readonly final class Foo
{
    public function __construct(
        /** @var list<Bar> */
        public array $items,
    ) {
    }
}
```

```php
namespace Acme;

readonly final class Bar
{
    public function __construct(
        public string $value,
    ) {
    }
}
```

```php
(new MapperBuilder())->mapper()->map(Foo::class, Source::array([
    'items' => [
        ['value' => 'baz'],
    ],
]));
```

Stacktrace:
```
LogicException: Namespace not found.
vendor/cuyz/valinor/src/Utility/Reflection/TokenParser.php:129
vendor/cuyz/valinor/src/Utility/Reflection/TokenParser.php:49
vendor/cuyz/valinor/src/Utility/Reflection/PhpParser.php:61
vendor/cuyz/valinor/src/Utility/Reflection/PhpParser.php:30
vendor/cuyz/valinor/src/Type/Parser/Lexer/AliasLexer.php:65
vendor/cuyz/valinor/src/Type/Parser/Lexer/AliasLexer.php:39
vendor/cuyz/valinor/src/Type/Parser/Lexer/AliasLexer.php:28
vendor/cuyz/valinor/src/Type/Parser/Lexer/TypeAliasLexer.php:27
vendor/cuyz/valinor/src/Type/Parser/LexingParser.php:30
vendor/cuyz/valinor/src/Type/Parser/LexingParser.php:29
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionTypeResolver.php:90
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionTypeResolver.php:65
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionTypeResolver.php:30
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionPropertyDefinitionBuilder.php:26
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionClassDefinitionRepository.php:81
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionClassDefinitionRepository.php:77
vendor/cuyz/valinor/src/Definition/Repository/Reflection/ReflectionClassDefinitionRepository.php:65
[...]

PR #446, opened 2023-10-23
base namespaces are T_STRING when tokenized

robchett:
Not happy with the fix, it feels wrong, but it does appear to be an issue with PHPs internal TokenParser and not this package

https://3v4l.org/OQ37v

### DIFF
--- a/qa/PHPStan/Extension/ApiAndInternalAnnotationCheck.php
+++ b/qa/PHPStan/Extension/ApiAndInternalAnnotationCheck.php
@@ -34,7 +34,8 @@ final class ApiAndInternalAnnotationCheck implements Rule
             return [];
         }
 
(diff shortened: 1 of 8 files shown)


robchett:
Fixes https://github.com/CuyZ/Valinor/issues/394

romm:
Hi @robchett, thanks for the fix, it looks reasonable to me considering the issue.

Could you please provide a non-regression test for this?

robchett:
99% sure that's not the best way to test this, the specific namespace requirements make it a bit tricky. Happy to revisit if you have a suggestion for a better way to do this.

robchett:
Hi @romm, is anything else needed for this PR?

romm:
Hi @robchett, just changed things a bit but overall everything was fine by me. Thanks for the fix submission!

PR #455, opened 2023-12-16
Fix union scalar with object not supported

vdauchy:
This fix aims to solve this case which is not currently supported:

```
class Foo {
    public string|Bar $value;
} 

class Bar {}
```

Without this a `CannotResolveTypeFromUnion` is throwed.


romm:
Hi @vdauchy thanks a lot for the contribution. I've merged #376 because it had more tests for these usecases, but I did put your fix in `UnionNodeBuilder` that was better IMO (also, made you co-author of the fix).

PR #456, opened 2023-12-16
Feature: Allow deactivation of flattening of single property objects.

vdauchy:
This allow to disable the behavior described in the documentation: https://valinor.cuyz.io/1.7/usage/object-construction/#class-with-a-single-value

**Context:**

When using valinor on API payload, json can be return like this:
```
{
    "items" : ["foo", "bar", "baz"]
}
```

Currently this would be flattened.

The issue is that this payload was designed this way to evolve... and could change to:

```
{
    "items" : ["foo", "bar", "baz"],
    "page": 1
}
```

In that case the flattening will not operate and so the mapping will crash.

With this feature used with `allowSuperfluousKeys` alow a greater stability of the code on payload changes that shall be backward compatible.

### DIFF
--- a/src/Library/Container.php
+++ b/src/Library/Container.php
@@ -123,7 +123,8 @@ final class Container
                     $this->get(ClassDefinitionRepository::class),
                     $this->get(ObjectBuilderFactory::class),
                     $this->get(ObjectNodeBuilder::class),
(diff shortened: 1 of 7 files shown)


vdauchy:
Looks like this is solving a issue that I do not reproduce on latest version :thinking:

PR #465, opened 2024-01-08
fix: allow callable type to be compiled

### LINKED ISSUE #466: Callable Types are not cachable (anymore?)
Not sure if this is related but I do now get the following exception:

`CuyZ\Valinor\Definition\Repository\Cache\Compiler\Exception\TypeCannotBeCompiled` with 

> The type `CuyZ\Valinor\Type\Types\CallableType` cannot be compiled.

That issue did not appear in 1.8.0 (maybe due to all tests failing regarding the issue fixed here) and did not appear in 1.7.0 either.
I will try to dive deeper into this and provide a dedicated issue for that problem.

Potential fix is provided in #465

romm:
Fixes #466

### DIFF
--- a/src/Definition/Repository/Cache/Compiler/TypeCompiler.php
+++ b/src/Definition/Repository/Cache/Compiler/TypeCompiler.php
@@ -9,6 +9,7 @@ use CuyZ\Valinor\Type\Type;
 use CuyZ\Valinor\Type\Types\ArrayKeyType;
 use CuyZ\Valinor\Type\Types\ArrayType;
 use CuyZ\Valinor\Type\Types\BooleanValueType;
+use CuyZ\Valinor\Type\Types\CallableType;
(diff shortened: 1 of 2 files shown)


boesing:
This patch actually works as expected.
Somehow, it tried to cache `uasort` method of `ArrayObject` (which is part of one of the objects we try to create) and that led to the actual problem.

Thanks!

romm:
Great! Is your application running smoothly on v1.8.1 with this patch?

boesing:
Yeah, its not in production. I just executed unit tests and found that problem. Unfortunately, I am unable to verify if the whole application works as of now but most probably will do that in the next days.

romm:
Alright, let' go for it then. Keep me updated if you find any other issue!

romm:
@boesing released as [1.8.2](https://github.com/CuyZ/Valinor/releases/tag/1.8.2), enjoy! 🚀

PR #479, opened 2024-02-08
misc: refactor file system cache to improve performance

romm:
Some functionalities were not used and would decrease overall performance, those have been removed.


danog:
Well actually, we were using the compiled PHP cache :)

PR #487, opened 2024-03-05
Failing test case: Valinor cannot understand `array{k: T|list<T>}`

Ocramius:
In this test, Valinor is being challenged to understand de-serialization of a given hashmap when given a type `array{a-key: T|list<T>}`.

Specifically, at the time of this writing, Valinor only picks `T` during de-serialization, and seems to completely ignore the `list<T>` portion of the data.

This kind of de-serialization is relevant when dealing with XML structures, where an element could be singular or plural depending on context:

```xml
<Invoices>
  <Invoice>
    <!-- in this example, `Item` is singular, and needs to be handled as `Item` -->
    <Item><Price>123</Price></Item>
  </Invoice>
  <Invoice>
    <!-- in this example, `Item` is plural, and needs to be handled as `list<Item>` -->
    <Item><Price>456</Price></Item>
    <Item><Price>789</Price></Item>
  </Invoice>
</Invoices>
```

Initially discovered by @Tigerman55

PR #488, opened 2024-03-05
Failing test: constructors with input `array{key: T}` and `array{key: list<T>}` considered colliding

Ocramius:
In this test, Valinor considers two disjoint input types as colliding.

Similar to https://github.com/CuyZ/Valinor/pull/487 ( ffe0f0f21290ff6469da5c723172b9afc6aeba7d ), this scenario was detected while trying to map multiple constructors for XML structures that may present different data depending on singular/plural entries found:

```xml
<Invoices>
  <Invoice>
    <!-- in this example, `Item` is singular, and needs to be handled as `Item` -->
    <Item><Price>123</Price></Item>
  </Invoice>
  <Invoice>
    <!-- in this example, `Item` is plural, and needs to be handled as `list<Item>` -->
    <Item><Price>456</Price></Item>
    <Item><Price>789</Price></Item>
  </Invoice>
</Invoices>
```

In #487, we attempted to map a single constructor using `array{foo: T|list<T>}`, while in this patch, we found the issue because we attempted to attack the problem by declaring separate constructors that would work on `array{foo: T}` and `array{foo: list<T>}` disjointly, but failed to do so due to aggressive collision detection logic.

Initially discovered by @Tigerman55

### DIFF
--- a/tests/Integration/Mapping/ConstructorRegistrationMappingTest.php
+++ b/tests/Integration/Mapping/ConstructorRegistrationMappingTest.php
@@ -633,6 +633,33 @@ final class ConstructorRegistrationMappingTest extends IntegrationTestCase
             ->map(stdClass::class, []);
     }
 
+    public function test_non_intersecting_hashmap_type_constructors_do_not_lead_to_collisions(): void
+    {
+        $mapper = $this->mapperBuilder()
+            ->registerConstructor(
+                /** @param array{key: SimpleObject} $input */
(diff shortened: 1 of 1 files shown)


romm:
Just rebased on `master`, after merging #552, which was not an easy task!

Enjoy this new "feature" in the upcoming release. 😊

Ocramius:
Nice!

Issue #358, opened 2023-03-08
Top-level union with allowSuperfluousKeys and overlapping types behaves unpredictably


TimWolla:
Consider the following example:

```php
<?php

use CuyZ\Valinor\MapperBuilder;

require('vendor/autoload.php');

class Foo
{
    public function __construct(
        private readonly string $foo,
        private readonly string $baz,
    ) {
    }
}

class Bar
{
    public function __construct(
        private readonly string $bar,
        private readonly string $baz,
    ) {
    }
}

class Any
{
    public function __construct()
    {
    }
}

try {
    $payload = (new MapperBuilder())
        ->allowSuperfluousKeys()
        ->mapper()
        ->map(
            "Bar | Foo | Any",
            [
                "foo" => "foo",
                "baz" => "baz",
                "bar" => "bar",
            ]
        );

    var_dump($payload);
} catch (\CuyZ\Valinor\Mapper\MappingError $error) {
    $messages = \CuyZ\Valinor\Mapper\Tree\Message\Messages::flattenFromNode(
        $error->node()
    );

    foreach ($messages as $message) {
        echo $message;
    }
}
```

The input is valid for each of the three types in the union, due to allowing superfluous keys. It currently emits the fairly non-descript error:

> Invalid value array{foo: 'foo', baz: 'baz', bar: 'bar'}.

When I remove any of the two property in `Bar` it consistently maps into `Foo`. If I add a `$foo` property to `Bar`, it consistently maps into `Bar`. If I remove both `foo` and `bar` from the input it consistently maps into `Any`.

I would expect Valinor to behave consistently with overlapping types in an union. Either the input should always be rejected if multiple union cases match, or (any this would be my preference) the first matching case should be used.

My real-world use case is mapping OAuth 1 response query strings. The same route should both handle the two possible response cases („Success“ and „Error“) and a fallback that accepts anything and starts the OAuth flow (the „Any“ type in the example above). This would allow me to match each of the three cases by checking `$payload instanceof Success` etc, effectively acting as a tagged union.

Because I cannot guarantee that the query string is completely clean (e.g. due to google analytics tracking parameters), I am allowing superfluous keys. During testing to make sure that the `Any` type only matches if no other type matches, I stumbled across this issue.

romm:
Hi @TimWolla, sorry for the late answer.

I tried to understand your usecase but I think the "Foo"/"Bar" naming doesn't help. From the example you posted, it seems to me that the results are correct:

```php
[
[...]

PR #493, opened 2024-03-12
feat: improve union type narrowing during mapping

### LINKED ISSUE #356: allowSuperfluousKeys and union do not work together
Consider the following example:

```php
<?php

$payload = (new MapperBuilder())
    ->allowSuperfluousKeys()
    ->mapper()
    ->map(
        <<<'EOT'
            array {
                status: 200,
                foo: array {
                    bar: string,
                },
            } | array {
                status: 400,
                error: string,
            }
            EOT,
        Source::json(<<<'EOT'
                {
                    "status": 200,
                    "foo": {
                        "bar": "xxx"
                    }
                }
                EOT)
    );

var_dump($payload);
```

My schema is a partial API response, because I do not care about additional keys. This first example validates properly: There are no additional keys. However if I add one:

```php
<?php

$payload = (new MapperBuilder())
    ->allowSuperfluousKeys()
    ->mapper()
    ->map(
        <<<'EOT'
            array {
                status: 200,
                foo: array {
                    bar: string,
                },
            } | array {
                status: 400,
                error: string,
            }
            EOT,
        Source::json(<<<'EOT'
                {
                    "status": 200,
                    "foo": {
                        "bar": "xxx",
                        "yyy": "zzz"
                    }
                }
                EOT)
    );

var_dump($payload);
```

… the response fails to validate:

> Fatal error: Uncaught CuyZ\Valinor\Mapper\TypeTreeMapperError: Could not map type `array{status: 200, foo: array{bar: string}}|array{status: 400, error: string}`. An error occurred at path *root*: Value object(CuyZ\Valinor\Mapper\Source\Source) does not match any of `array{status: 200, foo: array{bar: string}}`, `array{status: 400, error: string}`. in /pwd/src/Mapper/TypeTreeMapper.php on line 29

It works properly if I remove the top-level union:

```php
<?php

$payload = (new MapperBuilder())
    ->allowSuperfluousKeys()
    ->mapper()
    ->map(
        <<<'EOT'
            array {
                status: 200,
                foo: array {
                    bar: string,
                },
            }
            EOT,
        Source::json(<<<'EOT'
                {
                    "status": 200,
                    "foo": {
                        "bar": "xxx",
                        "yyy": "zzz"
                    }
                }
                EOT)
    );

var_dump($payload);
```

### LINKED ISSUE #395: Union of scalar and array of classes does not work
Hi, when trying to map union of scalar and array of classes, i get error:
```
CuyZ\Valinor\Mapper\TypeTreeMapperError: Could not map type `Foo`. An error occurred at path items: Invalid value array{0: array{…}}.
```
```php
(new MapperBuilder())->mapper()->map(Foo::class, Source::array([
    'items' => [
        [
            'id' => 1,
            'value' => 'baz',
        ],
    ],
]));
```

```php
readonly final class Foo
{
    public function __construct(
        /** @var string|list<Bar> */
        public string|array $items,
    ) {
    }
}
```

```php
readonly final class Bar
{
    public function __construct(
        public int $id,
        public string $value,
    ) {
    }
}
```

When trying as source the string value (first type of the union) it works as expected. 
```php
(new MapperBuilder())->mapper()->map(Foo::class, Source::array([
    'items' => 'some value',
]));
```

Currently my workaround is to not map it to array of objects but to union of string and array shape and do the mapping later separately which works but it is not as comfortable as it could be.

romm:
The algorithm used by the mapper to narrow a union type has been greatly improved, and should cover more edge-cases that would previously prevent the mapper from performing well.

If an interface, a class or a shaped array is matched by the input, it will take precedence over arrays or scalars.

```php
    ->mapper()
    ->map(
        signature: 'array<int>|' . Color::class,
        source: [
            'red' => 255,
            'green' => 128,
            'blue' => 64,
        ],
    ); // Returns an instance of `Color`
```

When superfluous keys are allowed, if the input matches several interfaces, classes or shaped array, the one with the most children node will be prioritized, as it is considered the most specific type:

```php
    ->allowSuperfluousKeys()
    ->mapper()
    ->map(
        // Even if the first shaped array matches the input, the second one is
        // used because it's more specific.
        signature: 'array{foo: int}|array{foo: int, bar: int}',
        source: [
            'foo' => 42,
            'bar' => 1337,
        ],
    );
```

If the input matches several types within the union, a collision will occur and cause the mapper to fail:

```php
    ->mapper()
    ->map(
        // Even if the first shaped array matches the input, the second one is
        // used because it's more specific.
        signature: 'array{red: int, green: int, blue: int}|' . Color::class,
        source: [
            'red' => 255,
            'green' => 128,
            'blue' => 64,
        ],
    );

// ⚠️ Invalid value array{red: 255, green: 128, blue: 64}, it matches at
//    least two types from union.
```

Fixes #395
Fixes #356
Closes #487
Closes #495

### DIFF
--- a/docs/pages/usage/type-reference.md
+++ b/docs/pages/usage/type-reference.md
@@ -153,10 +153,10 @@ final class SomeClass
         private array $unionInsideArray,
         
         /** @var int|true */
-        private int|bool $unionWithLiteralTrueType;
+        private int|bool $unionWithLiteralTrueType,
         
         /** @var int|false */
-        private int|bool $unionWithLiteralFalseType;
+        private int|bool $unionWithLiteralFalseType,
         
         /** @var 404.42|1337.42 */
         private float $unionOfFloatValues,
--- a/src/Mapper/Object/FilteredObjectBuilder.php
+++ b/src/Mapper/Object/FilteredObjectBuilder.php
@@ -16,17 +16,23 @@ final class FilteredObjectBuilder implements ObjectBuilder
 {
     private ObjectBuilder $delegate;
 
-    private Arguments $arguments;
-
-    public function __construct(mixed $source, ObjectBuilder ...$builders)
+    private function __construct(mixed $source, ObjectBuilder ...$builders)
     {
         $this->delegate = $this->filterBuilder($source, ...$builders);
-        $this->arguments = $this->delegate->describeArguments();
+    }
+
+    public static function from(mixed $source, ObjectBuilder ...$builders): ObjectBuilder
+    {
+        if (count($builders) === 1) {
+            return $builders[0];
+        }
+
+        return new self($source, ...$builders);
     }
 
     public function describeArguments(): Arguments
     {
-        return $this->arguments;
+        return $this->delegate->describeArguments();
     }
 
     public function build(array $arguments): object
--- a/src/Mapper/Tree/Builder/CasterProxyNodeBuilder.php
+++ b/src/Mapper/Tree/Builder/CasterProxyNodeBuilder.php
@@ -5,6 +5,10 @@ declare(strict_types=1);
 namespace CuyZ\Valinor\Mapper\Tree\Builder;
 
 use CuyZ\Valinor\Mapper\Tree\Shell;
+use CuyZ\Valinor\Type\CompositeTraversableType;
+use CuyZ\Valinor\Type\Type;
+use CuyZ\Valinor\Type\Types\ShapedArrayType;
+use CuyZ\Valinor\Type\Types\UnionType;
 
 /** @internal */
 final class CasterProxyNodeBuilder implements NodeBuilder
@@ -16,11 +20,28 @@ final class CasterProxyNodeBuilder implements NodeBuilder
         if ($shell->hasValue()) {
             $value = $shell->value();
 
-            if ($shell->type()->accepts($value)) {
+            if ($this->typeAcceptsValue($shell->type(), $value)) {
                 return TreeNode::leaf($shell, $value);
             }
         }
 
         return $this->delegate->build($shell, $rootBuilder);
+    }
+
+    private function typeAcceptsValue(Type $type, mixed $value): bool
+    {
+        if ($type instanceof UnionType) {
+            foreach ($type->types() as $subType) {
+                if ($this->typeAcceptsValue($subType, $value)) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        return ! $type instanceof CompositeTraversableType
+            && ! $type instanceof ShapedArrayType
+            && $type->accepts($value);
     }
 }
--- a/src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
+++ b/src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CuyZ\Valinor\Mapper\Tree\Exception;
+
+use CuyZ\Valinor\Mapper\Tree\Message\ErrorMessage;
+use CuyZ\Valinor\Mapper\Tree\Message\HasParameters;
+use CuyZ\Valinor\Type\Types\UnionType;
+use CuyZ\Valinor\Utility\String\StringFormatter;
+use CuyZ\Valinor\Utility\TypeHelper;
+use RuntimeException;
+
+use function array_map;
+use function implode;
+
+/** @internal */
+final class TooManyResolvedTypesFromUnion extends RuntimeException implements ErrorMessage, HasParameters
+{
+    private string $body;
+
+    /** @var array<string, string> */
+    private array $parameters;
+
+    public function __construct(UnionType $unionType)
+    {
+        $this->parameters = [
+            'allowed_types' => implode(
+                ', ',
+                array_map(TypeHelper::dump(...), $unionType->types())
+            ),
+        ];
+
+        $this->body = TypeHelper::containsObject($unionType)
+            ? 'Invalid value {source_value}, it matches at least two types from union.'
+            : 'Invalid value {source_value}, it matches at least two types from {allowed_types}.';
+
+        parent::__construct(StringFormatter::for($this), 1710262975);
+    }
+
+    public function body(): string
+    {
+        return $this->body;
+    }
+
+    public function parameters(): array
+    {
+        return $this->parameters;
+    }
+}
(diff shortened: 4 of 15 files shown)


docs/pages/usage/type-reference.md
```diff
@@ -153,10 +153,10 @@ final class SomeClass
         private array $unionInsideArray,
         
         /** @var int|true */
-        private int|bool $unionWithLiteralTrueType;
+        private int|bool $unionWithLiteralTrueType,
         
         /** @var int|false */
-        private int|bool $unionWithLiteralFalseType;
+        private int|bool $unionWithLiteralFalseType,
```
Ocramius, line 159: Are the commas intentional here?

src/Mapper/Object/FilteredObjectBuilder.php
```diff
@@ -16,17 +16,23 @@ final class FilteredObjectBuilder implements ObjectBuilder
...
-    private Arguments $arguments;
-
-    public function __construct(mixed $source, ObjectBuilder ...$builders)
+    private function __construct(mixed $source, ObjectBuilder ...$builders)
     {
         $this->delegate = $this->filterBuilder($source, ...$builders);
-        $this->arguments = $this->delegate->describeArguments();
+    }
+
+    public static function from(mixed $source, ObjectBuilder ...$builders): ObjectBuilder
```
Ocramius, line 24: s/from/mix, perhaps? You seem to aggregate multiple builders?

src/Mapper/Tree/Builder/CasterProxyNodeBuilder.php
```diff
@@ -16,11 +20,28 @@ public function build(Shell $shell, RootNodeBuilder $rootBuilder): TreeNode
...
+                if ($this->typeAcceptsValue($subType, $value)) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        return ! $type instanceof CompositeTraversableType
+            && ! $type instanceof ShapedArrayType
```
Ocramius, line 44: Excluded on purpose? :thinking:

Ocramius: From a reader's perspective, it's not 100% clear why these checks are in place: shouldn't `CompositeTraversableType` and `ShapedArrayType` also have `accepts()`?

src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->parameters = [
+            'allowed_types' => implode(
+                ', ',
+                array_map(TypeHelper::dump(...), $unionType->types())
+            ),
+        ];
+
+        $this->body = TypeHelper::containsObject($unionType)
+            ? 'Invalid value {source_value}, it matches at least two types from union.'
```
Ocramius, line 35: Nice! We should probably add "couldn't take a decision" or such?

src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
```diff
@@ -0,0 +1,50 @@
...
+            : 'Invalid value {source_value}, it matches at least two types from {allowed_types}.';
```
Ocramius, line 36: s/at least/two or more

That's just because "at least" sounds like we need need to match more

Ocramius (approved):
LGTM, but please consider rebasing on top of #487 to preserve research/history

TimWolla:
Possibly related issues of mine you might want to look into: #356, #358.

docs/pages/usage/type-reference.md
romm: Yes! While reading the doc I noticed this old typo (these are promoted properties inside a constructor).

src/Mapper/Object/FilteredObjectBuilder.php
romm: It actually filters builders depending on the input, so that only one remains at the end. I added this static constructor to check if there is only one builder, in which case it can be returned as is and does not need the whole filtering process.

src/Mapper/Tree/Builder/CasterProxyNodeBuilder.php
romm: This is a tricky one, and I'm still not convinced this class is at the correct place.

The goal was to save resources by checking if the input is accepted by the current node's type, before trying casting. It led to issues during development of this PR, so I adapted it, but I think I need to find a better way to check whether the input is valid or not. I wrote it in my internal notes, I'll try to improve it in the future.

src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
romm: Thanks!

src/Mapper/Tree/Exception/TooManyResolvedTypesFromUnion.php
romm: What sentence would you have in mind?

FYI the `TypeHelper::containsObject` is there because right now the system cannot display a meaningful representation of an object constructor. This is because an object might have several constructors, for instance when taking the [`Color` class from the docs](https://valinor.cuyz.io/latest/how-to/use-custom-object-constructors/#manually-registering-a-constructor), its representation should be `non-empty-string|array{red: int<0, 255>, green: int<0, 255>, blue: int<0, 255>}`. I have plans to improve it, but this will be in a separate PR.

Ocramius:
```diff
-'Invalid value {source_value}, it matches at least two types from union.'
-'Invalid value {source_value}, it matches at two or more types in an union: cannot take a decision.'
```

This would already be a good improvement :)

tests/Integration/Mapping/UnionMappingTest.php
```diff
@@ -94,6 +94,18 @@ public static function union_mapping_works_properly_data_provider(): iterable
             'assertion' => fn (mixed $result) => self::assertSame(['key' => [42, 1337]], $result),
         ];
 
+        yield 'shaped array representing http response with status 200' => [
+            'type' => "array{status: 200, data: array{text: string}} | array{status: 400, error: string}",
+            'source' => ['status' => 200, 'data' => ['text' => 'foo']],
```
TimWolla, line 99: Note: The test case in my issue was specifically about allowing superfluous fields, because I don't care about extra fields in the response, i.e. this:

```
'source' => ['status' => 200, 'data' => ['text' => 'foo', 'extra' => 'bar']],
```

romm: Yeah, I [edited my message already](https://github.com/CuyZ/Valinor/issues/356#issuecomment-2002506064). Just pushed a new test case that should match!

TimWolla: Awesome, thank you!

romm:
There we go! 🎉

PR #495, opened 2024-03-14
Failing test case: `''|array{foo: int}` if superfluous key exists

Tigerman55:
Found an issue where valinor does not appear to interpret a union with an empty string and array when the array being mapped contains a superfluous key. I was also able to reproduce this bug using `string|array{foo:int}`.


romm:
Thanks for the report! You're definitely lucky, people have been complaining for this kind of errors for a long time and I just decided to fix it with #493. 😊

Tigerman55:
> Thanks for the report! You're definitely lucky, people have been complaining for this kind of errors for a long time and I just decided to fix it with #493. 😊

Very cool! 😎

Issue #427, opened 2023-09-29
Unneeded calls to the autoloader


TimWolla:
Please consider the following code:

```php
<?php

namespace NS;

use CuyZ\Valinor\MapperBuilder;

require('vendor/autoload.php');

\spl_autoload_register(function (string $class) {
    echo "L: ", $class, PHP_EOL;
});

final class Target
{
    /**
     * @param   array{
     *              foo?: string,
     *              bar?: int,
     *              baz?: array<string, int>,
     *          } $foo
     * @param   array<string,string> $bar 
     */
    public function __construct(
        public readonly array $foo,
        public readonly array $bar,
    )
    {
    }
}

try {
    $payload = (new MapperBuilder())
        ->mapper()
        ->map(
            Target::class,
            [
                "bar" => [],
                "foo" => [
                    "foo" => "foo",
                ]
            ]
        );

    var_dump($payload);
} catch (\CuyZ\Valinor\Mapper\MappingError $error) {
    $messages = \CuyZ\Valinor\Mapper\Tree\Message\Messages::flattenFromNode(
        $error->node()
    );

    foreach ($messages as $message) {
        echo $message, PHP_EOL;
    }
}
```

When running this code, the following output is emitted:

```
L: NS\array
L: NS\array
L: NS\foo
L: NS\foo
L: foo
L: foo
L: foo
L: NS\string
L: NS\string
L: NS\bar
L: NS\bar
L: bar
L: bar
L: bar
L: NS\int
L: NS\int
L: NS\baz
L: NS\baz
L: baz
L: baz
L: baz
L: NS\array
L: NS\array
L: NS\string
L: NS\string
L: NS\int
L: NS\int
L: NS\array
L: NS\array
L: NS\string
L: NS\string
L: NS\string
L: NS\string
L: NS\array
L: NS\array
L: NS\foo
L: NS\foo
L: foo
L: foo
L: foo
L: NS\string
L: NS\string
L: NS\bar
L: NS\bar
L: bar
L: bar
L: bar
L: NS\int
L: NS\int
L: NS\baz
L: NS\baz
L: baz
L: baz
L: baz
L: NS\array
L: NS\array
L: NS\string
L: NS\string
L: NS\int
L: NS\int
L: NS\array
L: NS\array
L: NS\string
L: NS\string
L: NS\string
L: NS\string
/pwd/test.php:44:
class NS\Target#146 (2) {
  public readonly array $foo =>
  array(1) {
    'foo' =>
[...]

PR #496, opened 2024-03-15
misc: refactor generic types parsing and checking

romm:
This refactor is needed for an upcoming change in the types parsing system, which will allow to distinguish native types and types coming from doc-blocks.


TimWolla:
Possibly of interest here: #427.

PR #512, opened 2024-03-31
`JsonNormalizer` accepting `json_encode` flags

### LINKED ISSUE #476: `Normalizer`: `float` precision
Hey there,

as we already had [issues](https://github.com/CuyZ/Valinor/issues/360) in valinor regarding `float` and `int` strict comparison, I wanted to bring this topic up.

I am not sure how different languages handle this, but I would assume that `int` can be a `float` as well, however, to prevent that, we sometimes used `JSON_PRESERVE_ZERO_FRACTION`.

Valinors `JsonNormalizer` is just casting floats to string which drops the zero fraction and converts floats to `int` in case they do not have decimals.

Should we either add a way to enable this behavior or is this ignored on purpose?
Current workaround is to use the array normalizer and `json_encode` on our own in combination with `JSON_PRESERVE_ZERO_FRACTION`.

boesing:
As discussed in #476, this PR introduces a new `JsonNormalizer#withOptions` method accepting the following `JSON_*` constant values as an integer bitmask:

- `JSON_HEX_QUOT`
- `JSON_HEX_TAG`
- `JSON_HEX_AMP`
- `JSON_HEX_APOS`
- `JSON_INVALID_UTF8_IGNORE`
- `JSON_INVALID_UTF8_SUBSTITUTE`
- `JSON_NUMERIC_CHECK`
- `JSON_PRESERVE_ZERO_FRACTION`
- `JSON_UNESCAPED_LINE_TERMINATORS`
- `JSON_UNESCAPED_SLASHES`
- `JSON_UNESCAPED_UNICODE`

Since we should always enable `JSON_THROW_ON_ERROR`, I haven't added that one to the possible constants as it is enabled by-default.

Closes #476


boesing:
Hm, need to update my GPG key 🤔

boesing:
FYI: This adds additional time to the static analysis checks due to the `int-mask-of`.

romm:
@boesing excellent work, thank you for the contribution!

I pushed some changes, I'll let you review them and comment if needed.

When it's ready, let's ship it! 🚀

docs/pages/serialization/normalizing-json.md
```diff
@@ -72,8 +72,54 @@ $normalizer = (new \CuyZ\Valinor\MapperBuilder())
...
+
+$lowerManhattanAsJson = $normalizer->normalize(
+    new \My\App\Coordinates(
+        longitude: 40.7128,
+        latitude: -74.0000
+    )
+);
+
+// `$lowerManhattanAsJson` is a valid JSON string representing the data:
+// {"longitude":"40.7128","latitude":-74.0000}
```
boesing, line 99: hm, somehow the longitude got string here. need to change this

src/Normalizer/JsonNormalizer.php
```diff
@@ -15,16 +14,91 @@
...
+     *
+     * $lowerManhattanAsJson = $normalizer->normalize(
+     *     new \My\App\Coordinates(
+     *         longitude: 40.7128,
+     *         latitude: -74.0000
+     *     )
+     * );
+     *
+     * // `$lowerManhattanAsJson` is a valid JSON string representing the data:
+     * // {"longitude":"40.7128","latitude":-74.0000}
```
boesing, line 95: longitude shouldnt be string here as well

docs/pages/serialization/normalizing-json.md
romm: Good catch!

romm:
@boesing just updated after https://github.com/CuyZ/Valinor/issues/476#issuecomment-2029823690

Waiting for your review before merging!

boesing:
LGTM! Thanks for talking over. ☺️

src/Normalizer/Formatter/JsonFormatter.php
```diff
@@ -33,7 +36,11 @@ public function format(mixed $value): void
         } elseif (is_bool($value)) {
             $this->write($value ? 'true' : 'false');
         } elseif (is_scalar($value)) {
-            $this->write(json_encode($value, JSON_THROW_ON_ERROR));
+            /**
+             * @phpstan-ignore-next-line / Due to the new json encoding options feature, it is not possible to let SA
+             *                             tools understand that JSON_THROW_ON_ERROR is always set.
+             */
+            $this->write(json_encode($value, $this->jsonEncodingOptions));
```
TimWolla, line 43: Should it be this instead (untested)?
```suggestion
            $this->write(json_encode($value, $this->jsonEncodingOptions | JSON_THROW_ON_ERROR));
```

romm: I think it's better (performance-wise) to disable the SA check than forcing it at runtime when it has been forced before.

src/Normalizer/Formatter/JsonFormatter.php
```diff
@@ -59,7 +66,9 @@ public function format(mixed $value): void
                 $isFirst = false;
 
                 if (! $isList) {
-                    $this->write('"' . $key . '":');
+                    $key = json_encode($key, $this->jsonEncodingOptions);
+
+                    $this->write($key . ':');
```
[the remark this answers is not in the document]
romm: Good catch!

I think the test is good as-is, furthermore we cannot access `json_last_error()` because `JSON_THROW_ON_ERROR` is forced at this time.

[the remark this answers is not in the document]
TimWolla: > I think the test is good as-is, furthermore we cannot access `json_last_error()` because `JSON_THROW_ON_ERROR` is forced at this time.

I mean it could also be something like this:

```
json_decode($jsonResult, flags: JSON_THROW_ON_ERROR);
```

and then the Exception would implicitly fail the test if the generated JSON is not actually JSON. I think it would be useful as a defense-in-depth check, because I can see someone copy-and-pasting the actual value `{0:"foo","b":"bar"}` into the test, without double-checking that it's actually valid JSON. The missing quotes are easy to miss.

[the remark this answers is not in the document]
romm: Good idea indeed, thanks! Just submitted #518, tell me if the solution seems ok to you.

PR #516, opened 2024-04-03
qa: Replace Reflection-helper related phpstan-ignore-line by phpstan-assert-if-true

TimWolla:
Related to #515.


romm:
Excellent, thank you!

PR #526, opened 2024-04-10
fix: skip node build for NullType when value is not null

NanoSector:
Version 1.11 revamped union types but also deleted a fast path present in 1.10. In our use cases with many large objects with many nullable properties, this caused execution times to nearly double, especially with XDebug enabled.

The following test case sees its execution time drop from ~6.7 seconds to ~3.6 seconds on my local machine with the included patch. Profiling the test sees a reduction of node build calls to nearly half the original value.

```php
<?php

declare(strict_types=1);

namespace CuyZ\Valinor\Tests\Integration\Mapping;

use CuyZ\Valinor\Mapper\MappingError;
use CuyZ\Valinor\Tests\Integration\IntegrationTestCase;

final class LotsOfObjectsMappingTest extends IntegrationTestCase
{
    public function test_can_map_lots_of_objects(): void
    {
        $objects = [];

        for ($i = 0; $i < 10000; ++$i) {
            $objects[] = [
                'valueA' => 'foo',
                'valueB' => 'bar',
                'valueC' => [
                    'valueA' => 'baz',
                    'valueB' => 'qux',
                ],
            ];
        }

        try {
            $result = $this->mapperBuilder()
                ->mapper()
                ->map(
                    'array<' . SomeClassWithNullableProperties::class . '>',
                    Source::array($objects)
                );
        } catch (MappingError $error) {
            $this->mappingFail($error);
        }

        self::assertCount(10000, $result);
    }
}

final class SomeClassWithNullableProperties
{
    public function __construct(
        public string                               $valueA,
        public string                               $valueB,
        public SomeClassWithNullableProperties|null $valueC = null,
        public SomeClassWithNullableProperties|null $valueD = null,
    ) {}
}
```

### DIFF
--- a/src/Mapper/Tree/Builder/UnionNodeBuilder.php
+++ b/src/Mapper/Tree/Builder/UnionNodeBuilder.php
@@ -13,6 +13,7 @@ use CuyZ\Valinor\Type\IntegerType;
 use CuyZ\Valinor\Type\ScalarType;
 use CuyZ\Valinor\Type\StringType;
 use CuyZ\Valinor\Type\Types\InterfaceType;
+use CuyZ\Valinor\Type\Types\NullType;
 use CuyZ\Valinor\Type\Types\ShapedArrayType;
 use CuyZ\Valinor\Type\Types\UnionType;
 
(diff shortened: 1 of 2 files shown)


romm:
Very good, thank you @NanoSector! I pushed some minor changes.

PR #527, opened 2024-04-12
feat: support normalizing `Ds\Collection`

simPod:
When I use transformers

```php
    ->registerTransformer(fn (\Ds\Set $v) => $v->toArray())
    ->registerTransformer(fn(Value $v) => $v->get())
    ->normalizer(\CuyZ\Valinor\Normalizer\Format::json())
```

My `\Ds\Set<Value>` is normalized to `array<Value>` but the Value is then ignored since the serialization is not recursive for the `Set`.

This fixes it.

Initially, I wanted to make this work for all `IteratorAggregate`s and `Iterator`s but with then `'iterable class'` case does not pass since it is designed to serialize its properties and not inner Iterator.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       php-version: '8.1'
-      php-extensions: yaml
+      php-extensions: ds,yaml
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
           - "8.3"
 
     env:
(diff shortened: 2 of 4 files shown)


romm:
Thanks @simPod! I've decided to change the behavior regarding iterable objects, which fixes the issue you had with `Ds\Collection` objects, or any other collection objects. 🙂

simPod:
Awesome!

romm:
@simPod Hi! Would you have a real-life example of why this feature would be needed? 😊

simPod:
E.g. serde of a ds map:

```php
use \Ds\Map;

/** @var Map<list<int>, string> */
$m = new Map();
$m->put([1], 'le value');
$m->put([1, 2], 'le value2');

$normalizer->normalize($m);
```

or just a simple generator `yield [1]=>'le value'`

PR #536, opened 2024-05-01
feat: add microseconds support to timestamp format

NanoSector:
Prior to this patch, this would require a custom constructor in the form of:

```php
static fn(float | int $timestamp): DateTimeImmutable => new
DateTimeImmutable(sprintf("@%d", $timestamp)),
```
This bypasses the datetime format support of Valinor entirely. This is required because the library does not support floats as valid DateTimeInterface input values.

This PR adds support for floats and registers timestamp.microseconds (U.u) as a valid default format.


romm:
Excellent! Thanks you @NanoSector.

PR #537, opened 2024-05-02
Add support for value-of<BackedEnum>

robchett:


robchett:
Added Integration tests too, raised an issue with my original work.

romm:
Thank you for you work @robchett! This will be available in next release.

romm (approved):

PR #544, opened 2024-07-05
misc: Replace `sha1()` by `hash('sha1', ...)`

TimWolla:
The dedicated `sha1()` function is proposed for deprecation with PHP 8.4: https://wiki.php.net/rfc/deprecations_php_8_4#deprecate_md5_sha1_md5_file_and_sha1_file

For a future change it might also be considered to move to a cheaper hash function. The xxHash family (available as `xxh*`) is faster and should be sufficiently collision resistant for use within the cache.


romm:
Hi @TimWolla thank you! 

> For a future change it might also be considered to move to a cheaper hash function.

Could we not use it already? What particular function(s) would work here?

TimWolla:
> Could we not use it already?

We probably can. For this PR I just proposed the minimally viable change, because I did not have the time to thoroughly investigate what replacement would be appropriate.

> What particular function(s) would work here?

As suggested in the commit message, the xxHash family might be appropriate. An alternative might be using `preg_replace()` to replace any special characters by an "escape sequence". Something like the following:

```
preg_replace('/[^a-zA-Z0-9]/', function ($char) {
  return sprintf('_%02x', ord($char));
});
```

This would be guaranteed to be collision resistant, because it's a bijection. It would also make the cache identifiers human-readable.

romm:
For future reference: https://github.com/symfony/symfony/pull/47094

romm:
Hello @TimWolla, thanks for the original work and explanations.

After reading a bit about this topic (check [this php.watch article](https://php.watch/versions/8.1/xxHash) for instance), I've decided to use the `xxh128` hash, as Valinor requires PHP 8.1+.

I also moved the hash logic into one place only.

I'll wait for your approbation before merging.

NB: squash commit message to be used:

```
misc: use `xxh128` hash algorithm for cache keys

This algorithm is way faster than the originally used `sha1`.

This commit also moves the hash logic into one place, saving even more
processing time. 

Reference: https://php.watch/versions/8.1/xxHash
```

romm:
I wouldn't be an expert without the community around this tool, composed by many excellent PHP developers. 😉 

Thanks again!

PR #547, opened 2024-07-17
fix(transformers): allow phpdoc for callables

simPod:
Currently, when I have some docblock for callable in Transformer, the `TransformerHasInvalidCallableParameter` is thrown. I want to allow more specific callable, phpstan requires it (`missingType.callable`).


romm:
Thank you @simPod!

Issue #542, opened 2024-05-19
`JsonNormalizer` creates array JSON when object is passed


boesing:
Hey there,

I wonder if it is possible to have valinor keeping the actual type of the object passed to `normalize`.

When I use the following code, I'd expect `{}` from being generated but instead, it returns `[]`:

```php
$normalizer = (new MapperBuilder())->normalizer(Format::json());
var_dump($normalizer->normalize(new stdClass));
// string(2) "[]"
```

I would not expect a type transition to an array just because I pass an object without properties.

simPod:
have you tried withOptions https://www.php.net/manual/en/json.constants.php#constant.json-force-object?

simPod:
Hm, not supported now. Maybe https://github.com/CuyZ/Valinor/issues/542

romm:
Thanks for the report @boesing, this should be fixed with #549.

PR #548, opened 2024-07-17
feat(json): allow normalizing stdClass to `{}`

simPod:
- allow `JSON_FORCE_OBJECT` flag

Resolves https://github.com/CuyZ/Valinor/issues/542


src/Normalizer/Formatter/JsonFormatter.php
```diff
@@ -42,6 +43,12 @@ public function format(mixed $value): void
              */
             $this->write(json_encode($value, $this->jsonEncodingOptions));
         } elseif (is_iterable($value)) {
+            if ($value === [] && $this->jsonEncodingOptions & JSON_FORCE_OBJECT) {
```
romm, line 46: If we support `JSON_FORCE_OBJECT` we should support it fully, not only for empty array.

romm: I've just pushed a change that handles the option correctly. Some tests are missing, still.

Also, this does not really solve what is described in #542: the `JSON_FORCE_OBJECT` should not be mandatory to have an empty `stdClass` normalized to `{}` instead of `[]`. I'll have to think about it.

But the `JSON_FORCE_OBJECT` looks interesting though. What's your opinion on this, @boesing?

boesing: yah, imho object should always remain object. I'd rather have that as default enabled but that might be a bc break - dunno.

romm: > yah, imho object should always remain object

Agreed; this PR would mostly be about lists that should be converted to JSON object, see: https://3v4l.org/qKB7g

Only tests are missing from this PR to be merged, if you got time feel free to do it, otherwise I'll do it at some point. 🙂

PR #551, opened 2024-08-08
Add PHP 8.4 compatibility

Slamdunk:
Fixes 2 issues:

```
CuyZ\Valinor\Mapper\Tree\Shell::child(): Implicitly marking parameter $attributes as nullable is deprecated, the explicit nullable type must be used instead
CuyZ\Valinor\Type\Types\ClassStringType::__construct(): Implicitly marking parameter $subType as nullable is deprecated, the explicit nullable type must be used instead
```

See https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

This PR **doesn't** add PHP 8.4 support, it just makes it runnable in PHP 8.4 without errors.

### DIFF
--- a/src/Mapper/Tree/Shell.php
+++ b/src/Mapper/Tree/Shell.php
@@ -53,7 +53,7 @@ final class Shell
         return (new self($settings, $type))->withValue($value);
     }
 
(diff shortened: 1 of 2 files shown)


romm:
Thank you @Slamdunk!

PR #555, opened 2024-09-02
fix: properly handle class sharing class name and namespace group name

### LINKED ISSUE #554: Class "Money\Money\Currency" does not exist
Hi, maybe I am doing something wrong but I get the following error:

```
Class "Money\Money\Currency" does not exist
```

reproducible example:

```php
<?php

declare(strict_types=1);

use Money\Currency;
use Money\Money;

require_once __DIR__ . '/vendor/autoload.php';

readonly class Transaction
{
    public Currency $currency;

    public string $amount;

    public function foobar(): Money
    {
        return new Money($this->amount, $this->currency);
    }
}

$mapper = (new MapperBuilder())
    ->allowSuperfluousKeys()
    ->mapper();

$mapper->map(
    Transaction::class,
    Source::array([
        'currency' => 'USD',
        'amount' => '100',
    ])
);
```

Any ideas on what I am doing wrong?

romm:
Fixes #554

### DIFF
--- a/src/Type/Parser/Factory/Specifications/AliasSpecification.php
+++ b/src/Type/Parser/Factory/Specifications/AliasSpecification.php
@@ -53,32 +53,27 @@ final class AliasSpecification implements TypeParserSpecification
 
     private function resolveAlias(string $symbol): string
     {
-        $alias = $symbol;
+        $aliases = PhpParser::parseUseStatements($this->reflection);
 
         $namespaceParts = explode('\\', $symbol);
-        $firstPart = array_shift($namespaceParts);
 
-        if ($firstPart) {
-            $alias = strtolower($firstPart);
+        $lastPart = strtolower(end($namespaceParts));
+
+        if (isset($aliases[$lastPart])) {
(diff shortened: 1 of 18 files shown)


src/Type/Parser/Factory/Specifications/AliasSpecification.php
```diff
@@ -64,17 +68,17 @@ private function resolveAlias(string $symbol): string
...
+        $fqcn = $aliases[$alias]['fqcn'];
 
-        $full = $aliases[$alias];
+        if ($namespaceParts === []) {
+            return $fqcn;
+        }
 
-        if (! empty($namespaceParts)) {
-            $full .= '\\' . implode('\\', $namespaceParts);
+        if (! $aliases[$alias]['isExplicitAlias']) {
```
stof, line 77: this looks wrong to me. The non-empty namespace parts must always be respected.

romm:
@stof right, I just gave it another look and could greatly simplify it. Thanks for noticing it, I guess I was tired the other day. 😅

Issue #558, opened 2024-09-07
Interface resolver


mastir:
Problem:

Currently we have 2 options: 
1) resolve interface by declaration in MapperBuilder (wich can resolve only one class as result)
2) resolve interface with strongly typed custom constructor

```php

interface Container{}

class BoxContainer{
    function __construct(int $width, int $height){}
}

class SquereContainer{
    function __construct(int $width, int $height){}
}

class CircleContainer{
    function __construct(int $radius){}
}
```

And there is no way to implement this with custom constructor. We could probably try to do somthing like

```php
class SomeContainer implements Container{
    public readonly $container;
    __construct(string $type, ?int $width, ?int $height, ?int $radius ){
        if ($type === 'box')$this->container = new BoxContainer($width, $height);
        if ($type === 'squere')$this->container =  new SquereContainer($width, $height);
        if ($type === 'circle')$this->container =  new CircleContainer($radius);
    }
}
```
And define it in MappginBuilder. But in my project i have 17 implementations with more then 200 posible arguments. Wich is totaly destroys any input validation.

Purposal:

Implement interface resolver, something like:
```php
$builder->withInterfaceResolver($resolver);

interface InterfaceResolver {
    public function resolve(string $interface): ?callable;
}

class CustomResolver implements InterfaceResolver {
    public function resolve(string $interface){
        if ($interface === 'Container'){
            return function($properies){
                if ($properties['type']==='box')return 'BoxContainer';
                if ($properties['type']==='squere')return 'SquereContainer';
                if ($properties['type']==='circle')return 'CircleContainer';
            }
        }
    }
}
```

PR #560, opened 2024-09-09
add dynamic interface resolution

mastir:
Made interface resolvers implementation for #558 . This implementation allows to add discriminator properties to normalized array and resove actual class based on descriminator props

Example:

```php
interface Container{}

class BoxContainer implements Container{
    function __construct(int $width, int $height){}
}

class SquereContainer implements Container{
    function __construct(int $width, int $height){}
}

class CircleContainer implements Container{
    function __construct(int $radius){}
}

class MyResolver(){
    public function resolve(string $interface, ?array $props) : ?string
    {
        if ($interface === 'Container'){
            return match($props['type']){
                'box' => BoxContainer::class,
                'squere' => SquereContainer::class,
                'circle' => CircleContainer::class,
            }
        }
    }
    
    public function getResolverProps(string $interface) : array
    {
        return ['type'];
    }
    
    public function transform(object $input, callable $next) : array
    {
         $result = $next($input);
         if ($input instanceOf ContainerInterface){
              $result['type'] = match(get_class($input)){
                 BoxContainer::class => 'box',
                SquereContainer::class => 'squere',
                CircleContainer::class => 'circle',
             }
         }
         return  $result;
    }
}

$mapperBuilder->withInterfaceResolver(new MyResolver());
$normalized  = $normalizer->normalize(new BoxContainer(3,5)); // ['type'=>'box', 'width'=>3, 'height'=>5]
$object = $mapper->map(Container::class, $normalized); //new BoxContainer(3,5)
```

This can partialy be done with infer / transform, but some validations faild due too extra fields and dynamic interface reading is not possible. In my project i use Attribute on interfaces to declare type mapping and discriminator field:

```php
#[TypedInterface([
    'box' => BoxContainer::class,
    'squere' => SquereContainer::class,
    'circle' => CircleContainer::class,
], 'conteiner_type')]
interface Container{
...

class InterfaceObjectResolver implements \CuyZ\Valinor\InterfaceResolver
{

    public function resolve(string $interface, ?array $props) : ?string
    {
        [$type_map, $discriminator] = $this->readAttributes($interface);
        if (!$type_map){
            return null;
        }
        return $type_map[$props[$discriminator]];
    }
...
```

### DIFF
--- a/src/InterfaceResolver.php
+++ b/src/InterfaceResolver.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace CuyZ\Valinor;
+
+interface InterfaceResolver
+{
+    public function resolve(string $interface, ?array $props) : ?string;
+
+    public function getResolverProps(string $interface) : array;
+
+    public function transform(object $input, callable $next) : array;
+}
--- a/src/InterfaceResolver.php
+++ b/src/InterfaceResolver.php
@@ -2,11 +2,24 @@
 
 namespace CuyZ\Valinor;
 
+/**
+ * @api
+ */
 interface InterfaceResolver
 {
-    public function resolve(string $interface, ?array $props) : ?string;
+    /**
+     * @param array<string,mixed>|null $props
+     * @return class-string|null
+     */
+    public function resolve(string $interface, ?array $props): ?string;
 
-    public function getResolverProps(string $interface) : array;
+    /**
+     * @return string[]
+     */
+    public function getResolverProps(string $interface): array;
 
-    public function transform(object $input, callable $next) : array;
+    /**
+     * @return array<string|int,mixed>
+     */
+    public function transform(object $input, callable $next): array;
 }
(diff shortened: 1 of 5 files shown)


romm:
This feature already exists in the library, see https://github.com/CuyZ/Valinor/issues/558#issuecomment-2358780243

Thanks for the effort, though! I'll have to think about making documentation more precise about it.

mastir:
Thanks for reply but its not solving real problem: Mapper is a service, services are loaded on startup. At the bootstrap we do not have interfaces, we cannot use functions with not yet existing classes and interfaces. Preloading wholeproject at bootstrap to implement interface resolution is not a solution.

So there is only 2 ways:
1) On every module loaded generate resoulver functions to infer into mapper builder, just in case we will ever need em later (wich is not in 99% cases). And have a lots of dead code in memory. Loose cache every time and recreate all bindings again and again.
2) add dynamic interface resolution. 

Maybe there is a better way to use Valinor with DI and dynmic class loading (like composer). Foor somthing like symfony,laravel,yii where all busines logic classes are not preloaded before services

PR #567, opened 2024-10-12
feat: add support for PHP 8.4

romm:
References:

- [ ] php/php-src/pull/16129
- [ ] vimeo/psalm/pull/10928


src/Definition/Repository/Reflection/ReflectionFunctionDefinitionRepository.php
```diff
@@ -83,7 +85,8 @@ public function for(callable $function): FunctionDefinition
      */
     private function signature(ReflectionFunction $reflection): string
     {
-        if (str_contains($reflection->name, '{closure}')) {
+        // PHP8.4 Only `str_starts_with($name, '{closure:')` is needed
```
TimWolla, line 88: With 8.2 you can just use [`ReflectionFunction::isAnonymous()`](https://www.php.net/manual/en/reflectionfunction.isanonymous.php).

nickvergessen:
I guess you don't feel comfortable to release a version without Psalm CI on your side, so you have a proof.
But would be helpful for your consumers already to be able to check and test things going forward :see_no_evil:

romm:
Hi @TimWolla sorry to poke you here, but as you fixed the `ReflectionFunction::inNamespace()` issue in https://github.com/php/php-src/pull/16129 I thought I could ask for some help. Hope you have time. 😊

It seems that even with the latest PHP 8.4-RC, we're still having an issue with [\CuyZ\Valinor\Tests\Integration\Mapping\Namespace\NamespacedInterfaceInferringTest::test_interface_inferred_from_same_namespace_as_file_runs_correctly](https://github.com/CuyZ/Valinor/blob/950395ec8d00695e3b311a101af19ea81074427e/tests/Integration/Mapping/Namespace/NamespacedInterfaceInferringTest.php#L13) ; the issue lies in [\CuyZ\Valinor\Type\Parser\Factory\Specifications\AliasSpecification::resolveNamespaced()](https://github.com/CuyZ/Valinor/blob/950395ec8d00695e3b311a101af19ea81074427e/src/Type/Parser/Factory/Specifications/AliasSpecification.php#L91): the closure is namespaced but the method still returns `false`.

Any idea what is wrong here? Thanks!

TimWolla:
@romm Closures are never considered namespaced with PHP 8.4+. Generally speaking you can't really rely on a specific closure naming, because Closures are anonymous and thus you would be relying on an implementation detail. From how the PHP code looked like, even the namespacing of Closures prior to 8.4 was an accident (they are just named `{closure}` and the namespace was implicitly added like for any other function). There weren't any tests either.

romm:
@TimWolla thanks for your answer.

The usecase here is to get the namespace of a closure so that short class names in docblock can be transformed to FQCN. For instance:

```php
namespace SimpleNamespace;

interface SomeInterface {}

final class ImplementationOne implements SomeInterface {}

final class ImplementationTwo implements SomeInterface {}

return (new MapperBuilder())
    ->infer(
        SomeInterface::class,
        /** @return class-string<ImplementationOne|ImplementationTwo> */
        static fn (string $type): string => match ($type) {
            'one' => ImplementationOne::class,
            'two' => ImplementationTwo::class,
            default => throw new \RuntimeException(),
        },
    )
    ->mapper()
    ->map(
        SomeInterface::class,
        ['type' => 'one'],
    );
```

The annotation `@return class-string<ImplementationOne|ImplementationTwo>` should understand that `ImplementationOne` should be fetched from namespace `SimpleNamespace`.

It seems strange to me that closures cannot be namespaced, is there any chance the old behavior could be reverted and tested? If not, would you have a suggestion on how to get this information? Using `PhpToken::tokenize` on the source file content and fetching the correct namespace?

TimWolla:
> It seems strange to me that closures cannot be namespaced

Namespaces are not a first class citizen in PHP. It's really just compiler assisted automated prefixing of any names. The `getNamespaceName()` methods in Reflection really do just that: Extract the part before the last backslash.

> is there any chance the old behavior could be reverted and tested?

Unfortunately not for PHP 8.4, it's much too late in the release cycle.

> If not, would you have a suggestion on how to get this information?

The effect on docblock parsing is an unfortunate side-effect of the improved closure naming that I did not anticipate when initially implementing it.

> Using PhpToken::tokenize on the source file content and fetching the correct namespace?

For any closure that is not a top-level closure you can strip the `{closure:` prefix and then find the last backslash. Anything in-between is the namespace. For top-level closures (i.e. closures without any surrounding function or method) you would need to tokenize the file, I'm afraid.

romm:
Thank you @TimWolla for the details. Tokenization was not that hard, and all this stuff gets cached anyway (if the cache is used), so that's not a big deal.

TimWolla:
@romm FWIW: The tokenization solution is not _quite_ correct in all cases, because the following is legal PHP code:

```
<?php
namespace Foo {
 use Bar\Bar;
 $closure = function () { var_dump(new Bar()); };
 class Foo { }
 var_dump($closure());
}

namespace Bar {
 use Foo\Foo;
 $closure = function () { var_dump(new Foo()); };
 class Bar { }
 var_dump($closure());
}

namespace {
 $closure = function () { };
}
```
```

File: src/Type/Parser/Factory/Specifications/AliasSpecification.php [hunk 2]
```diff
@@ -88,11 +88,17 @@ private function resolveNamespaced(string $symbol): string
             }
         }
 
-        if (! $reflection->inNamespace()) {
+        if ($reflection->inNamespace()) {
+            $namespace = $reflection->getNamespaceName();
+        } elseif ($reflection instanceof ReflectionFunction) {
```
TimWolla, line 93: You could guard this against closures to not unnecessarily tokenize the file for top-level functions in non-namespaced code.

PR #582, opened 2024-11-22
misc: remove the need to keep a reference to the parent node

Baptouuuu:
## Problem

The current implementation keeps a reference in each `Shell` to its parent in order to know if it's the root node and be able to generate the its path.

The problem with this design is that this creates circular references, making the job difficult for PHP to free these objects from memory.

## Solution

Since we build a child from the parent object we create the path by duplicating the path of the parent and adding the child name.


romm:
Thank you! 😊

PR #584, opened 2024-12-05
misc: fix PHP 8.4 deprecations

mpesari:
Implicitly nullable parameter types are deprecated in PHP 8.4:

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


romm:
Thanks!

PR #594, opened 2025-01-18
feat: introduce a benchmark system

Baptouuuu:
This commit introduce a benchmark system based on phpbench.

In essence for each new pull request it runs benchmarks on the master branch and then run the same benchmarks against the code from the pull request and compare them.

If the new code average execution time goes above 10% it will fail the pull request.

This commit only focuses into introducing this system and not create an exhaustive set of benchmarks. There are still a few benchmarks created in order to verify the system works.

It also adds a bit of documentation since the phpbench output can be hard to understand if not already familiar with it.

## Benchmarks

The `initialisationBuilder` and `initialisationMapper` benchmarks are here to have an indication of the time it takes to initialise Valinor. Since the mapper should be built only once in a project, the time of other benchmarks should deduct this initialisation time.

As said in the documentation there's a warmup run to _hide_ the autoload part. But it may interesting to not hide this time from the 2 benchmarks named above.

## Documentation

You may want to rework the documentation in order to fit the rest of the documentation style.

## Possible further work

In case the benchmark comparison fails, one needs to go look at the output to know what went wrong. It would be nice to display the output directly as a PR comment.

As a starting point it would be interesting to create benchmarks for each type supported by Valinor specified in this [documentation](https://valinor.cuyz.io/latest/usage/type-reference/). This would provide a good baseline. After that, new benchmarks should come from real use cases.

Both suggestion can be done later on in new Pull Requests.


Baptouuuu:
Didn't think of this but the new benchmark workflow can't work in this PR since there's no benchmark on `master` yet 🤦‍♂️.

Should I remove the workflow and create a new PR when this one will be merged into `master` ?

romm:
Thank you! 😊

> Should I remove the workflow and create a new PR when this one will be merged into `master` ?

Yeah let's do this.

romm:
Ok, waiting for the new PR! 🫡

PR #602, opened 2025-02-10
Allow option to use dynamic properties for specific classes

marcelthole:
Hey,
i have a special use case where i need to allow dynamic properties via `__get` and `__set` in a class.
But with the strict type handling this package will not allow this, because all parameters must be known already. 

So i created this PR for discussion if this solution would be fine for you.
The next idea could also be to check for the `#[\AllowDynamicProperties]` class attribute - But this will takle a little bit other problem, where you don't need the `__get` and `__set` at all.

### DIFF
--- a/docs/pages/usage/type-strictness-and-flexibility.md
+++ b/docs/pages/usage/type-strictness-and-flexibility.md
@@ -102,3 +102,40 @@ mapping.
         ]
     );
 ```
+
+## Allowing dynamic Properties
+
+In some cases it could be required to map dynamic properties if the class implements the magic `__get` and `__set` Methods.
+To archive this you should mark the classes with an Interface and all not explicit allowed values would be allowed.
+
+```php
+interface MagicGetterSetter {
+    public function __get(string $name): string|null;
+    public function __set(string $name, string $value): void;
+}
+
+class DynamicClass implements MagicGetterSetter {
+    /** @var array<string, string> */
+    public array $dynamicProperties = [];
+    public function __get(string $name): string|null { 
+        return $this->dynamicProperties[$name] ?? null; 
+    }
+    public function __set(string $name, string $value): void {
+        $this->dynamicProperties[$name] = $value;
+    }
+}
+
+(new \CuyZ\Valinor\MapperBuilder())
+    ->allowDynamicPropertiesFor(MagicGetterSetter::class)
+    ->mapper()
+    ->map(
+        DynamicClass::class,
+        [
+            'foo' => 'foo',
+            'bar' => 'baz',
+        ]
+    );
+
+// Result
+// DynamicClass::$dynamicProperties' = ['foo' => 'foo','bar' => 'baz']
+```
--- a/src/Library/Settings.php
+++ b/src/Library/Settings.php
@@ -50,6 +50,9 @@ final class Settings
 
     public bool $allowPermissiveTypes = false;
 
+    /** @var list<class-string> */
+    public array $allowDynamicPropertiesFor = [];
+
     /** @var callable(Throwable): ErrorMessage */
     public $exceptionFilter;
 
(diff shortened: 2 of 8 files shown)


marcelthole:
I have still one isse and need your help here. I would like to use also generics for this implementation
```php
/**
 * @template TValue
 * @extends \ArrayAccess<string, TValue>
 */
interface DynamicProperties extends \ArrayAccess {
    /** @return list<string> */
    public function getAdditionalPropertyKeys(): array;
    /** @return TValue */
    public function __get(string $name): mixed;
    /** @param TValue $value */
    public function __set(string $name, mixed $value): void;
    public function __unset(string $name): void;
    public function __isset(string $name): bool;
}
```

No i also want to use this strict types here and to validate the given values. Currently this implementation use the given type and i can't find a way to get the implements (or extends if i wrap it in an abstract class) `TValue` - or is it easier to get the type from the `__set` method signature? But this is without the `TValue` also only mixed

marcelthole:
@romm could you give me an Update for this PR? is this a feature this package could contain for the future?

romm:
Hi @marcelthole, sorry for the late answer and thanks for the contribution.

Before taking a deep look at the changes, I'd like to know more about your real usecase, that made you want this feature. Could you tell me more?

What seems to be missing is mostly a support of the `ArrayAccess`, I think the `allowDynamicPropertiesFor` part is unnecessary. As you pointed out, we also miss something in the code to handle the generic types of `ArrayAccess`, this is currently not possible (although it shouldn't be that hard to implement).

marcelthole:
Hey,
in my use case i have an OpenAPI Document that i want to map. 
This is an example class:

```php

/** @extends AdditionalProperties<string> */
class Info extends AdditionalProperties {
	public string $title;
	public null|string $summary = null;
	public null|string $description = null;
	public null|string $termsOfService = null;
	public null|Contact $contact = null;
	public null|License $license = null;
	public string $version;
}

/** 
 * @template TValue
 * @extends \ArrayAccess<string, TValue>
 */
abstract class AdditionalProperties extends \ArrayAccess
{
    /** @param array<string, TValue> $additionalProperties */
    private array $additionalProperties = [];
   // ...
}
```

The problem is, that the OpenAPI Spec file allows additional properties starting with `X-`. So i would like to have an feature that handle this case. I have a strict mapping for my known fields, but there could be additional fields in the Payload that i didn't know. With the extends annotations i could also say: I allow additional properties i didn't know but they all need to be a string.

By implementing a Tagging interface i could get this unknown additional properties only for these classes where i need it.

romm:
Hi @marcelthole, I'm closing this in favor of the following solution, made possible by attribute converters which were introduced in #673:

```php
<?php

declare(strict_types=1);

use CuyZ\Valinor\Mapper\AsConverter;

#[Attribute(Attribute::TARGET_CLASS), AsConverter]
final class AdditionalPropertyConverter
{
    /**
     * @param array<mixed> $value
     * @param callable(array<mixed>): object $next
     */
    public function map(array $value, callable $next): object
    {
        foreach ($value as $key => $item) {
            if (str_starts_with(strtolower($key), 'x-')) {
                $value['additionalProperties'][$key] = $item;
                unset($value[$key]);
            }
        }

        return $next($value);
    }
}

#[AdditionalPropertyConverter]
final class Info
{
    public string $title;

    public null|string $description = null;

    /** @var array<string, string> */
    public array $additionalProperties = [];
}

$result = (new \CuyZ\Valinor\MapperBuilder())
    ->mapper()
    ->map(Info::class, [
        'title' => 'Some title',
        'description' => 'Some description',
        'X-Some-Additional-Key' => 'Foo',
    ]);

$result->title === 'Some title';
$result->description === 'Some description';
$result->additionalProperties === ['X-Some-Additional-Key' => 'Foo'];
```

Hope this works well for you!

PR #603, opened 2025-02-17
allow non array-key like keys in iterables (DsMap)

simPod:
@romm Hi, a follow up of this 

https://github.com/CuyZ/Valinor/pull/527#issuecomment-2353783797

I have adapted the test in order to allow other key types in iterables but before I dive further I'd like your opinion on it. Thanks!

### DIFF
--- a/tests/Integration/Normalizer/NormalizerTest.php
+++ b/tests/Integration/Normalizer/NormalizerTest.php
@@ -62,12 +62,18 @@ final class NormalizerTest extends IntegrationTestCase
             $builder = $builder->registerTransformer($transformerAttribute);
         }
 
-        $arrayResult = $builder->normalizer(Format::array())->normalize($input);
+        if($input instanceof \Ds\Map){
+        } else {
+            $arrayResult = $builder->normalizer(Format::array())->normalize($input);
+        }
         $jsonNormalizer = $builder->normalizer(Format::json())->withOptions($jsonEncodingOptions);
 
         $jsonResult = $jsonNormalizer->normalize($input);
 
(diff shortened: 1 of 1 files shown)


tests/Integration/Normalizer/NormalizerTest.php
```diff
@@ -221,13 +227,17 @@ public static function normalize_basic_values_yields_expected_output_data_provid
...
+            $input->put(['foo'], 'foo');
+            $input->put(['bar'], 'bar');
+
             yield 'Ds Map' => [
-                'input' => new \Ds\Map(['foo' => 'foo', 'bar' => 'bar']),
+                'input' => $input,
                 'expected array' => [
-                    'foo' => 'foo',
-                    'bar' => 'bar',
+                    ['k' => ['foo'] , 'v' => 'foo'],
```
romm, line 237: Where are the `'k'` and `'v'` coming from? Is this some sort of consensus or an opinionated value?

romm:
Hi @simPod sorry for the late answer. I'm not sure this will be an easy task to handle this, but we'll see. I posted one question concerning the name of the normalized keys, before we go further.

tests/Integration/Normalizer/NormalizerTest.php
simPod: It is opinionated. IMO `key` and `value` are too long, especially for data serialization. Thus `k` `v`.

Maybe it could even be `[['foo'], 'foo']`, like an `Entry` is in js. 

key value pair, KVP.

simPod:
Hello Romain, sure thing.

PR #635, opened 2025-05-03
qa: Use runBare method now that runTest is made private

NanoSector:
PHPUnit commit
https://github.com/sebastianbergmann/phpunit/commit/8f0f8205e3a38675392da55afc573c6bb6a43f4e makes the `runTest` method used by the integration tests private. `runBare` however is still public, so we can use that.

Because `runBare` will also call pre-test and post-test hooks, we have to be careful not to end in an infinite loop.

Thought: Tests should potentially be rewritten to be wrapped in a closure instead, which runs the test with and without cache. This however would take a lot more time to implement than this quick fix. This would break the dependency on PHPUnit internal code. We could then add a post-test hook to verify a cache was used.

### DIFF
--- a/tests/Integration/IntegrationTestCase.php
+++ b/tests/Integration/IntegrationTestCase.php
@@ -38,14 +38,19 @@ abstract class IntegrationTestCase extends TestCase
             return;
         }
 
+        // If we are already running with a cache, avoid an endless loop by bailing out now.
+        if (isset($this->cacheToInject)) {
+            return;
+        }
+
         $cacheDir = sys_get_temp_dir() . DIRECTORY_SEPARATOR . bin2hex(random_bytes(16));
         $this->cacheToInject = new FileSystemCache($cacheDir);
 
         // First rerun of the test: the cache entries will be injected.
-        parent::runTest();
+        parent::runBare();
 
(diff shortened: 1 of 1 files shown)


tests/Integration/IntegrationTestCase.php
```diff
@@ -38,16 +38,21 @@ protected function tearDown(): void
...
         // First rerun of the test: the cache entries will be injected.
-        parent::runTest();
+        parent::runBare();
 
         // Second rerun of the test: the cache entries will be used and tested.
-        parent::runTest();
+        parent::runBare();
 
-        $this->cacheToInject->clear();
+        unset($this->cacheToInject);
```
romm, line 55: Any particular reason for this change? This removes the clearing of the populated cache entries, which we'd prefer to have.

romm:
Awesome, thank you! 🤗

tests/Integration/IntegrationTestCase.php
NanoSector: The thought behind this was proactively resetting the state of the cache object before the next test run since we're now in more 'global' PHPUnit state territory.

In practice this doesn't really change anything, so I've reverted the change :)

PR #648, opened 2025-06-06
feat: Add PHPUnit extension for pretty-printing MappingError instances

NanoSector:
This helps with debugging MappingErrors thrown during test execution. Sometimes these can be annoying to debug because Valinor only shows "A total of 2 errors were thrown".

The extension works by overriding the TestCase `transformException` method through a trait, which means this is an opt-in feature for tests. It collects them in a singleton class and then uses the PHPUnit extension to display them in a formatted table at the end of the test run.

image

Implementing it is a matter of using the `CollectValinorMappingErrors` trait in your TestCase class and registering the
`PrettyPrintMappingErrorsExtension` extension class in the phpunit configuration file. A PR to update the documentation will be made separately.

This does require symfony/console for its table printing capabilities, copying that logic into the library seemed excessive. In order to not have a hard dependency on it, it is put in the suggest array in composer.json and we are using a class_exists check to see if it is installed.

Developed by: Social Deal (@socialdeal)\

### DIFF
--- a/.github/workflows/static-analysis-plugins.yml
+++ b/.github/workflows/static-analysis-plugins.yml
@@ -47,6 +47,14 @@ jobs:
         with:
           working-directory: "qa/StaticAnalysis/phpstan/v2"
 
+      - uses: "ramsey/composer-install@v2"
+        with:
+          working-directory: "qa/StaticAnalysis/phpunit/v10"
+
+      - uses: "ramsey/composer-install@v2"
+        with:
+          working-directory: "qa/StaticAnalysis/phpunit/v11"
+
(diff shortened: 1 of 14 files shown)


NanoSector:
PHPStan is (rightfully) complaining that the `CollectValinorMappingErrors` trait is unused. I could throw it in some random test or suppress it as the 'static analysis' tests cover its usage; what is wanted here?

romm:
Hi @NanoSector just wanted to say that I did not forget you. Most efforts lately have been done towards the 2.0 release. I do not have much personal time to work on Valinor lately but I'll take a look at this PR ASAP. 😊

Thanks!

PR #652, opened 2025-06-18
update vimeo/psalm, infection/infection

staabm:
upgrade to newer versions, as I want to compare latest infection-dev version vs. latest stable.

we can't upgrade even more as this would require PHP 8.2+

---

infection runtime before the update (v0.27.11) (4965 mutations)
> Time: 3m 22s. Memory: 1.39GB. Threads: 6

after the update (v0.29.8) (4967 mutations)
> Time: 2m 4s. Memory: 0.50GB. Threads: 6

vs. dev-master d820fec (4516 mutations)
> Time: 1m 42s. Memory: 0.47GB. Threads: 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,13 +21,13 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^10.5",
-        "infection/infection": "^0.27",
+        "infection/infection": "^0.29",
         "phpstan/phpstan": "^2.0",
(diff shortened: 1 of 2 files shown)


maks-rafalko:
Very nice observations! I was wondering what was the reason of such a great performance improvement, looks like this PR, see 2 links to Infection Playground: https://github.com/infection/infection/pull/2003#issuecomment-2463233754

Cool!

PR #734, opened 2025-10-14
fix: handle default value retrieval for properties

eliashaeussler:
Calling `ReflectionProperty::getDefaultValue()` on a property without a default value is deprecated since PHP v8.5[^1]. With this change, the method is now only called if the property actually contains a default value.

[^1]: https://github.com/php/php-src/pull/19457


romm:
Woops! Completely missed this PR 😅 

Thanks!

Issue #733, opened 2025-10-14
bottleneck in `toShapedArray`


staabm:
we just measured a use-case in which we map lots of data.

in this use-case we found the following bottleneck in valinor.
maybe the callgraph is useful, so you it can be optimized?

<img width="225" height="375" alt="Image" src="https://github.com/user-attachments/assets/016af505-64bd-4ac8-a671-b3054404db05" />

staabm:
> The first one probably can be improved with memoization or something similar.

my gut feeling is, getting rid of the array_map() and using regular foreach will already be a improvement.
I will measure that.

> so saving them in memory should not be an issue

I did not yet have a memory problem. can only tell more if we try to memoize it and check against the current workload :)

romm:
Everything has been merged in `master`, would you run a new benchmark to see the final improvements in this area? 🙂

PR #735, opened 2025-10-16
Micro-optimize Arguments

staabm:
refs https://github.com/CuyZ/Valinor/issues/733

grafik

less function call overhead, as we remove millions of array_map() calls

grafik


src/Mapper/Object/Arguments.php
```diff
@@ -28,13 +28,15 @@
 final class Arguments implements IteratorAggregate, Countable
 {
     /** @var array<string, Argument> */
-    private array $arguments = [];
+    private readonly array $arguments;
```
staabm: while here, more immutability is a good thing IMO

PR #738, opened 2025-10-20
chore: use memoization for type dumping

romm:
See #733

@staabm I'd love to have a benchmark feedback on this one as well 😊

### DIFF
--- a/src/Type/Dumper/TypeDumper.php
+++ b/src/Type/Dumper/TypeDumper.php
@@ -28,14 +28,22 @@ use function usort;
 /** @internal */
 final class TypeDumper
 {
+    /** @var array<string, string> */
+    private array $dumpedTypes = [];
+
     public function __construct(
-        private readonly ClassDefinitionRepository $classDefinitionRepository,
-        private readonly ObjectBuilderFactory $objectBuilderFactory,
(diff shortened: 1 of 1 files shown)


src/Type/Dumper/TypeDumper.php
```diff
@@ -28,14 +28,22 @@
...
-        private readonly ObjectBuilderFactory $objectBuilderFactory,
+        private ClassDefinitionRepository $classDefinitionRepository,
+        private ObjectBuilderFactory $objectBuilderFactory,
         private InterfaceInferringContainer $interfaceInferringContainer,
         private FunctionsContainer $constructors,
     ) {}
 
     public function dump(Type $type): string
+    {
+        return $this->dumpedTypes[$type->toString()] ??= $this->dumpAndQuote($type);
```
staabm: couldn't this cache be more effective if we would not have calls circumventing it (e.g. direct calls to `doDump()` as we have them in this file)?

romm: I'm sorry, I don't get what you mean? 😅

staabm: within this class here we have a lot of calls to `doDump()`. these do not benefit from this cache here and might do unnecessary duplicate work (maybe I am wrong)

PR #757, opened 2025-11-13
fix: simplify namespace resolution

### LINKED ISSUE #756: Using short class names in docblock does not work, in closures that are imported into a class
```php
use \Implementations\ImplementationA;

$mapperBuilder = new MapperBuilder()
            ->allowSuperfluousKeys()
            ->allowScalarValueCasting()
            ->infer(
                SomeInterface::class,
                /** @return class-string<ImplementationA|\implementations\ImplementationB> */
                fn($data): string => !isset($data['containerId']) ? ImplementationA::class : ImplementationB::class
            )
            ->allowPermissiveTypes()
        ;
```

The `ImplementationA` type will be unresolvable. 

This happens for a specific scenario:
1. The file itself has no namespace
2. The file is included via `require` inside a class that has a namespace

In `PHPParser::fetchUseStatements` the namespace is for the closure is looked up:
```php
    private static function fetchUseStatements(ReflectionClass|ReflectionFunction|ReflectionMethod $reflection): array
    {
        if ($reflection instanceof ReflectionMethod) {
            $namespaceName = $reflection->getDeclaringClass()->getNamespaceName();
        } elseif ($reflection instanceof ReflectionFunction && $reflection->getClosureScopeClass()) {
            $namespaceName = $reflection->getClosureScopeClass()->getNamespaceName();
        } else {
            $namespaceName = $reflection->getNamespaceName();
        }

        $content = self::getFileContent($reflection);

        if ($content === null) {
            return [];
        }

        return (new TokenParser($content))->parseUseStatements($namespaceName);
    }

```

But in the case of a configuration file that is included (possibly from within a class as is the case in my Codeception test environment). The namespace is resolved incorrectly.

I'll finish this bug report tomorrow.

SamMousa:
Fixes #756

### DIFF
--- a/src/Utility/Reflection/PhpParser.php
+++ b/src/Utility/Reflection/PhpParser.php
@@ -57,20 +57,23 @@ final class PhpParser
             return [];
         }
 
+        $tokenParser = new TokenParser($content);
+
         if ($reflection instanceof ReflectionMethod) {
             $namespaceName = $reflection->getDeclaringClass()->getNamespaceName();
-        } elseif ($reflection instanceof ReflectionFunction && $reflection->isStatic() && $reflection->getClosureScopeClass()) {
-            $namespaceName = $reflection->getClosureScopeClass()->getNamespaceName();
+        } elseif ($reflection instanceof ReflectionFunction) {
+            $namespaceName = $tokenParser->getNamespace();
         } else {
             $namespaceName = $reflection->getNamespaceName();
         }
 
-        return (new TokenParser($content))->parseUseStatements($namespaceName);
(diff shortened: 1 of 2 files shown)


romm:
Hi @SamMousa thanks for the report and the PR! There was indeed a change in how PHP 8.4 returns a function reflection's namespace, see https://github.com/php/php-src/issues/18070 for instance. My first attempt to fix it (in ad354dd639279eb84f2145be8068113a8346f15a) was run on PHP 8.2 so the result was different 😅 

Anyway, I adapted your code a bit, now it works perfectly. Thanks for the new test cases!

PR #795, opened 2026-03-24
fix: Allow passing pure-closures to NormalizerBuilder::registerTransformer

thePanz:
This will solve PHPStan errors when passing Closure to registerTransformer

Ref: https://phpstan.org/writing-php-code/phpdoc-types#callables

### DIFF
--- a/src/NormalizerBuilder.php
+++ b/src/NormalizerBuilder.php
@@ -173,7 +173,7 @@ final class NormalizerBuilder
      * {@see https://en.wikipedia.org/wiki/Pure_function}
      *
      * @pure
-     * @param pure-callable|class-string $transformer
+     * @param pure-callable|pure-Closure|class-string $transformer
      */
     public function registerTransformer(callable|string $transformer, int $priority = 0): self
     {


romm:
Hi @thePanz, I thought `@pure-callable` would include `@pure-closure` 🤔 

Maybe I'm wrong, but do you have an example where it fails?

thePanz:
I get this error:
image

And that's the class:
```php
final readonly class ConvertKeysToSnakeCase implements NormalizerBuilderConfigurator
{
    public function configureNormalizerBuilder(NormalizerBuilder $builder): NormalizerBuilder
    {
        return $builder->registerTransformer(static function (object $object, callable $next): mixed {
            $result = $next();

            if (! is_array($result)) {
                return $result;
            }

            $snakeCased = [];

            foreach ($result as $key => $value) {
                $lcFirstKey = preg_replace('/[A-Z]/', '_$0', lcfirst($key));
                $newKey = strtolower($lcFirstKey ?? $key);

                $snakeCased[$newKey] = $value;
            }

            return $snakeCased;
        });
    }
}```

Maybe the fix is not really working, as for some phpstan-cache on my current project? :thinking:

romm:
Oh, I see. Then I doubt that adding `pure-closure` will change anything to the issue.

I gave some explanations [in the documentation](https://valinor-php.dev/latest/other/static-analysis/#about-mapper-and-normalizer-purity) about purity — TL;DR it is quite hard to handle pure markers in a codebase, so I give access to a PHPStan plugin that will silent those errors for you (this is enabled by default in Valinor's own codebase BTW).

I'm closing this, but thanks anyway 😊

PR #799, opened 2026-03-26
feat: introduce object to represent an HTTP request problem details

romm:


boesing:
I wonder why valinor needs these kind of details.
Yes, this is an RFC standard, yet it feels odd that this type of detail is held within valinor. 

I have seen that you are using it in the mezzio package and it might be useful in symfony bundle, yet this is not related to the core of valinor (i.e. normalization/deserialization).

I am not against this, it just feels like something which should reside in the respective package(s).

---

Oh and btw., the "errors" are actually not RFC. Only status, type, detail and title are.
Just wondering if this follows some framework specific details. For example in symfony, the errors are mapped differently than in mezzio.

romm:
Hi @boesing, thanks for your interest!

The work on this is to ease the use of RFC-7807 problems when using Valinor for [HTTP request mapping](https://valinor-php.dev/latest/how-to/map-http-request/) (you may have missed this new feature from not so long ago).

HTTP request mapping is one of the usages that makes most sense with Valinor, that's why I wanted it provided out of the box by the library. Problem details is a logical follow-up. 😊

> Just wondering if this follows some framework specific details. For example in symfony, the errors are mapped differently than in mezzio.

Actually this is more RFC-compliant than any-framework compliant, and I'm pretty sure we can find packages that do this kind of job for well-known frameworks. When this PR is merged, I will provide an implementation for Symfony and Mezzio bundles/packages to automatically make us of it.

PR #806, opened 2026-05-13
feat: add support for `key-of` type mapping

danog:


danog:
Ping @romm :)

romm:
Thank you @danog!

PR #807, opened 2026-05-14
feat: add support for shaped list type

danog:
Also includes the key-of implementation to avoid conflicts since key-of also applies to shaped lists


danog:
Ping @romm :)

romm:
Hi @danog, thanks for the submission. I did a bit of refactoring to use the same class to parse both (shaped) lists and arrays, to avoid a lot of duplicated logic which would have made the maintenance harder.

danog:
Thanks, looks okay!

romm:
Thank you @danog!

PR #808, opened 2026-05-18
Add support for traits in PhpParser

danog:


danog:
nvm this was misguided

PR #810, opened 2026-05-19
Add support for magic properties

danog:
Submitting this to fully upstream changes in the valinor fork of one of my clients.

While currently Valinor does not make use of the new `magicProperties` property, it is currently used by one of my clients to implement automatic type validation for their ORM, which relies exclusively on magic properties, specifically by taking all properties, compiling them via PropertyDefinitionCompiler::compile and storing them in a single static cache class, regenerated on every build.
 
While I know this is technically using an `@internal` API, the current caching options offered by Valinor are suboptimal in our case, where a simple static cache is enough , and much more performant than the alternatives.


romm:
Hi @danog, I'm not sure I want this to land in the codebase. I'll leave it open and try to spend some reflection on it when I have more time for OSS.

PR #824, opened 2026-07-18
bug: failing to infer generic type on anonymous class constructor, when imported types used

Ocramius:


tests/Integration/Mapping/GenericTypesInAnonymousClassConstructorTest.php
```diff
@@ -0,0 +1,82 @@
...
+                ]
+            );
+        
+        self::assertEquals(
+            new DummyPaginable('abc123', [new User('ocramius')]),
+            $result->wrapped,
+        );
+    }
+
+    public function test_deserializes_generic_type_with_fqcn_references(): void
```
Ocramius, line 56: this test passes, the previous fails

Ocramius: hmm, somehow, the test passes when run with the entire test suite? 

Running the single test locally seems to fail it for me :thinking:

romm:
Overridden by #825. Thank you for the contribution!

PR #831, opened 2026-07-31
test: add regression tests for scalar value casting with `asRoot`

devnix:
Route and query parameters from an HTTP request are always strings, and the mapper enables scalar value casting for them, as documented:

> Route and query parameter values coming from an HTTP request are typically strings. The mapper automatically handles scalar value casting for these parameters: a string `"42"` will be properly mapped to an `int` parameter.

The existing `asRoot` test cases pass native values (`'anotherQueryParameter' => 42`); that path is never exercised with the strings a real framework provides. This adds two cases: one mapping all query parameters to a shaped array, one to an object alongside a `#[FromRoute]` parameter.

This is not a bug report: `master` already behaves correctly. The bug existed on `2.5.1` and was fixed incidentally by 578bd58 (`chore: refactor HTTP request mapping`), a commit not yet in any release. These tests pin the fix down.

Reproduction on `2.5.1`:

```php
$request = new HttpRequest(queryParameters: ['page' => '42']);

// Works.
(new MapperBuilder())->argumentsMapper()->mapArguments(
    fn (#[FromQuery] int $page) => [],
    $request,
); // => ['page' => 42]

// Fails.
(new MapperBuilder())->argumentsMapper()->mapArguments(
    /** @param array{page: int} $query */
    fn (#[FromQuery(asRoot: true)] array $query) => [],
    $request,
); // => MappingError: An error occurred at path page: Value '42' is not a valid integer.
```

These tests guard the fix. Reverting the change in `HttpRequestNodeBuilder` (excluding the `asRoot` key from `allowScalarValueCastingForChildren()`) makes both fail with `Value '42' is not a valid integer`; the other 56 cases in the file pass.

Happy to adjust naming or placement if you prefer them elsewhere.


romm:
Thank you!

PR #841, opened 2026-09-18
Use StructArmed on QA

samsonasik:
Hi, this PR add StructArmed to QA

https://github.com/boundwize/structarmed

It is a configurable PHP architecture guards with built in presets (psr4, mvc, and so on).

It already used by PHP frameworks:

| Project          | Configuration                                                                                |
| ---------------- | -------------------------------------------------------------------------------------------- |
| CakePHP          | https://github.com/cakephp/cakephp/blob/5.x/structarmed.php               |
| CodeIgniter 4    | https://github.com/codeigniter4/CodeIgniter4/blob/develop/structarmed.php |
| Spiral Framework | https://github.com/spiral/framework/blob/master/structarmed.php           |
| Mezzio Framework | https://github.com/mezzio/mezzio/blob/3.29.x/structarmed.php           |

This PR apply `StructArmed` with register:

- PSR4, CodeQuality preset. The CodeQuality preset apply `MustBeStaticAnonymousFunctionRule` to make use of `static` in anonymous function when possible, applied wtih:

```bash
vendor/bin/structarmed analyze --fix
```

- I also enable ruleset so it show more ability of layering setup to define which layers each layer is allowed to depend on.

Current applied `Source` is set to `src` and `qa` to easier to review.

- Register to Github workflow.

I also clean up no longer part of set rector config rule and re-run it, also fix phpstan notice on latest update.


samsonasik:
All green 🎉 Ready for review/merge 👍

romm:
Hi @samsonasik thanks for the PR. I'm not familiar with the tool, although it looks interesting. 

Some questions/remarks:

- I'm a bit familiar with [Deptrac](https://github.com/deptrac/deptrac), what would be the main differences with this tool?
- Is there a benefit to use the "CodeQuality" preset over PHP-CS-Fixer/Rector? We already have two linters in this project, I'd prefer to stick with these ones and not add a new one (although I see the value in adding layer rules) — in which case the static functions changes should be done by one of the existing tools.
- Does [the current ruleset](https://github.com/CuyZ/Valinor/pull/841/changes#diff-6c750795fe8e9ac3fc3936402fef25d2ca69ca1d01ef0f16c2ad8b27cde4b041R56-R92) look correct to you? Should we do some changes? The goal would be to later detect layers that get mixed up, so we'd better think it properly first.
- Looks like you have updated a lot of dev dependencies while installing the package, I'd prefer the update to go in a separate PR (or at least an atomic commit).

Thank you 😊

samsonasik:
Hi @romm,

1. It has similar feature on ruleset part, but faster. 

Here performance tested on `Spiral Framework` with 2215 files only on 0.6 second

spiral-fast

Here a list of characteristic `StructArmed` has:

| Area                     | StructArmed                                                                                                                                                             |
| ------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Configuration            | Native PHP configuration                                                                                                                                                |
| Getting started          | Ready-made presets, such as PSR-4, with layers, layer patterns, and rulesets added gradually as needed                                                                  |
| Preset customization     | Registered preset rules can be overridden, replaced, or skipped                                                                                                         |
| PHPUnit integration      | Can run as a PHPUnit extension                                                                                                                                          |
| Platform support         | Tested on Windows, macOS, and Linux                                                                                                                                     |
| Performance              | Optimized parallel analysis with optional caching                                                                                                                       |
| Extensibility            | Supports custom rules and custom fixers                                                                                                                                 |
| Automatic fixes          | Custom rules can implement `FixableInterface` and support `--fix`                                                                                                       |
| Parser/fixer integration | Custom rules and fixers can use `php-parser`, [JsonRecast](https://github.com/boundwize/jsonrecast), or their own implementation                                        |
| Dependencies             | Small set of direct runtime dependencies: [composer.json](https://github.com/boundwize/structarmed/blob/927f8c16de216ea46089e75a8d47b7cce07ba3b3/composer.json#L31-L34) |
| Tests                    | 100% test coverage                                                                                                                                                      |

You can read online documentation at https://boundwize.github.io/structarmed/

2. Rector is deprecating rules, eg:

- https://github.com/rectorphp/rector-src/pull/8092
- https://github.com/rectorphp/rector-src/pull/7569

that's why `StructArmed` has rules to cover that with `CodeQuality` preset: 

- `MustBeStaticAnonymousFunctionRule`
- `LargeNumericLiteralMustUseSeparatorRule`

More rules that deprecated on Rector side can be added in the future on `StructArmed` if make sense.

3. The ruleset is based on directory path, the inheritance of ruleset has `+LayerName` so it can Inheriting Allowed Layers, see documentation:

https://boundwize.github.io/structarmed/layers-and-rulesets/#inheriting-allowed-layers

if you think you need to be clean up first for starter, I can clean up them so we can re-structure later.

4. I can extract into separate PRs if you want:

    - Fix rector dependencies and deprecated rules in the config and re-run it (because PHPStan used by Rector, PHPStan fix needs apply as well)    
    - Setup minimal StructArmed config with initial preset, eg: `PSR4` preset.

wdyt?

romm:
> Here performance tested on Spiral Framework with 2215 files only on 0.6 second

Awesome, thanks.

> Rector is deprecating rules, eg:

Thanks for the clarification, in this case could you update the [composer check](https://github.com/CuyZ/Valinor/blob/master/composer.json#L50-L61) and [composer fix](https://github.com/CuyZ/Valinor/blob/master/composer.json#L65-L69) commands as well, please?

> if you think you need to be clean up first for starter, I can clean up them so we can re-structure later.

Not really sure what you mean there, sorry 😅 — you'd lower the rules to improve them later?

> I can extract into separate PRs if you want:

Yes, let's go for it. :)

samsonasik:
@romm what I mean is just use the inital psr4 preset, eg:

```php
<?php

declare(strict_types=1);

use Boundwize\StructArmed\Architecture;
use Boundwize\StructArmed\Preset\Preset;

return Architecture::define()
    // apply on qa and src directory for now
    // for ease gradual changes
    ->layer('Source', ['qa', 'src'])
    ->withPresets(Preset::PSR4());
```

but that mean no change yet to showoff the code quality rules :)

samsonasik:
See PR for bump Rector and PHPStan:

- https://github.com/CuyZ/Valinor/pull/842

samsonasik:
See PR for just initial use of StructArmed with only PSR4 preset:

- https://github.com/CuyZ/Valinor/pull/843

samsonasik:
handled on other PR

PR #843, opened 2026-09-21
qa: install architecture guard package `boundwize/structarmed`

samsonasik:
@romm here initial use of StructArmed with PSR4 Preset on QA.

- https://github.com/CuyZ/Valinor/pull/841

This apply to Source layer that PSR4 preset scans: `qa` and `src` directory to ease for gradual updates.

Per I write at https://github.com/CuyZ/Valinor/pull/841#issuecomment-5762188294, there is nothing to showoff to just PSR4 preset on `qa` and `src`, I can apply on `tests` directory, but there are many multiple classes in single file which can be discussed later whether they are on purpose.

PR #844, opened 2026-09-21
qa: Set dependency ruleset for StructArmed

samsonasik:
@romm continue of PR:

- https://github.com/CuyZ/Valinor/pull/843

This PR set dependency ruleset. Here are some explanations:

- setup layers that can be used in the ruleset
- layer that can be defined per directory set via `layer()`
- layer that can need exclusion if parent dir and current dir is different layer, use of `layerPattern()` with 3rd arg as excluded pattern.
- use of `+LayerName` syntax sugar to show inheriting allowed Layers to merge a layer’s allowed dependencies into another layer’s allowed list, so it can simplified:

```diff
'Definition' => ['Utility', 'UtilityType', 'Type', 'Attributes', 'FeatureAttribute'],

// ...
-'DefinitionCache' => ['Definition', 'Utility', 'UtilityType', 'Type', 'Attributes', 'FeatureAttribute', 'Cache'],
+'DefinitionCache' => ['+Definition', 'Cache'],
```

so it auto expand without repeated copy paste allowed layers.


samsonasik:
@romm ready to merge 👍

samsonasik:
@romm I've bumped to structarmed ^0.18 and change `layerPattern()` to `layer()` with 3rd argument for simplify config, no more regex in config 👍 

Ready for merge 👍

samsonasik:
@romm I've fixed phpstan notice after composer update. Ready to merge 👍

